PR #4, opened 2016-08-09
Feature/chart component

fungjj92:
screen shot 2016-08-09 at 12 02 00 pm

Does:
- Sidebar with real buttons that populate the chart section whoo!
- Populate chart section with active indicators
- Draw d3 charts with mock data

Since icons still don't work, I gave the export and options buttons temporary characters just to show up.

Outstanding functionality:
- Responsiveness of d3 charts (currently static rendering)
- Style, labels, real data for d3 charts


src/assets/css/chart.css
```diff
@@ -0,0 +1,72 @@
+.charts {
```
CloudNiner: Shouldn't the chart css go into it's component folder and be included via `styleUrls: []`?

src/assets/css/sidebar.css
```diff
@@ -0,0 +1,40 @@
+sidebar {
```
CloudNiner: same thing here as above?

src/ccl/charts/chart.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  providers: [ChartService],
+  directives: [LineGraph],
+  styleUrls: [
+  '../../assets/css/chart.css',
+  ],
+  templateUrl: './chart.template.html'
+})
+export class Chart {
+    // Pull active chart data
+    public chartService: ChartService;
```
CloudNiner: From what I've seen, this is an unnecessary declaration, if in the constructor you do:
```
constructor(private chartService: ChartService) { }
```

src/ccl/charts/chart.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    private chartData;
+
+    makeCharts() {
+      this.getChartData.subscribe(data=> {
+        this.chartData = data;
+      });
+    }
+
+    constructor(@Inject(ChartService) chartService:ChartService){
+      this.chartList = chartService.get();
```
CloudNiner: Nothing other than injection should happen in the constructor. Everything your controller needs to do should happen in `ngOnInit()`:

https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html

src/ccl/ccl.component.ts
```diff
@@ -3,16 +3,19 @@
  * App Component
  */
 import { Component, ViewEncapsulation } from '@angular/core';
+import { Sidebar } from './sidebar/sidebar.component';
+import { Chart } from './charts/chart.component';
```
CloudNiner: Since you added an index file for this folder, you should be able to just do `import { Chart } from './charts/';`

src/ccl/charts/line-graph.component.ts
```diff
@@ -0,0 +1,119 @@
...
+  selector: 'line-graph',
+  encapsulation: ViewEncapsulation.None,
+  template: `<ng-content></ng-content>`,
+  inputs: [ 'data', 'indicator' ],
+  styleUrls: [ '../../assets/css/chart.css']
+})
+
+export class LineGraph {
+
+  public data: Object;
```
CloudNiner: It would be nice to take advantage of typescript here and give this a type, something like `ChartData`? Or `LineGraphData`? This way its more clear to the developer what the input for this component is, and typescript can validate it at compile time.

src/ccl/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+ * Sidebar Component
+ * Populates sidebar with indicators and triggers adding charts
+ */
+
+@Component({
+  selector: 'sidebar',
+  encapsulation: ViewEncapsulation.None,
+  providers: [IndicatorsService, ChartService],
```
CloudNiner: Watch out. Whenever you add a provider to a component via this configuration, it instantiates a new instance of that provider at this level. So, if you did the same thing elsewhere in the app (which it looks like you're doing), you'd get two instances of the same provider. This can be good and bad depending on what you want. Bad if the provider in question is intended to be a singleton. This can cause subtle bugs, and is better explained here:

https://angular.io/docs/ts/latest/guide/hierarchical-dependency-injection.html#!#component-injectors

Read through to the end of the page. Might need to read the whole section for context. If this is confusing, I'm happy to explain tomorrow.

src/ccl/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    // Pull data to populate sidebar
+    constructor(private chartService: ChartService, private indicatorService: IndicatorsService){};
+
+    // Set up click event handlers
+    onIndicatorClicked(indicator) {
+        this.chartService.addChart(indicator);
+    }
+
+    ngOnInit() {
```
CloudNiner: Does this actually get called? You should also import `OnInit` and extend the class with it for bookkeeping either way.

src/ccl/charts/chart.component.ts
```diff
@@ -0,0 +1,41 @@
...
+import { LineGraph } from './line-graph.component';
+
+/*
+ * Chart component
+ * Management for the sections that appear in the chart container
+ */
+
+@Component({
+  selector: 'charts',
+  encapsulation: ViewEncapsulation.None,
```
CloudNiner: I'm not sure using this encapsulation method is in general a good idea. It feels awkward and breaks the idea that a component folder has all of the stuff necessary to use it. Does our webpack config need additional work to support the default 'emulated' mode?

src/ccl/charts/chart.template.html
```diff
@@ -0,0 +1,14 @@
+<div class="chart" *ngFor="let chart of chartList">
```
CloudNiner: This encapsulation feels weird. I would expect the `charts` directive to simply loop and display a list of `chart` components. That way the settings stuff can be encapsulated inside `chart` rather than having it live here. `chart` can still contain `line-graph`, I like that that lives separately from the chart's settings.

src/ccl/charts/chart.component.ts
fungjj92: You're right this method was not correct here. default emulated is fine. All the other components seem to require the None method. I don't think I understand Shadow DOM well even after reading though: http://blog.thoughtram.io/angular/2015/06/29/shadow-dom-strategies-in-angular2.html

src/ccl/sidebar/sidebar.component.ts
fungjj92: Didn't know about this!

src/ccl/ccl.component.ts
fungjj92: Doesnt work :X

src/ccl/sidebar/sidebar.component.ts
fungjj92: It does run: OnInit seems like fluff. We can continue with the 'extends OnInit' option.

src/ccl/charts/chart.template.html
fungjj92: Could you clarify? I'm not sure I understand the difference between the method you're explaining and what is here

fungjj92:
Sidebar buttons work now!!

src/ccl/sidebar/sidebar.component.ts
CloudNiner: Indeed. Just helpful type annotations then to make it easier to see what the class is doing. Also provides type checking in the case of you misspelling the method name or something and getting confused as to why the method isn't called.

src/ccl/ccl.component.ts
CloudNiner: Maybe the syntax I referred to in my previous comment is some sugar that the ng-cli provides. That has a concept of bundles. If it's something we have to set up, then this is fine.

src/ccl/charts/chart.template.html
CloudNiner: Yeah, mabe lets grab a conf room after scrum and draw out the view hierarchy on a whiteboard. Might help to see it.

src/ccl/charts/line-graph.component.ts
fungjj92: I'm not too up with typescript yet. Can type declaration, as in this case, basically be a comment? i.e. Object<ChartData>
As is there is no appropriate existing class

src/ccl/charts/charts-container.component.ts
```diff
@@ -0,0 +1,44 @@
...
+            </div>`
+})
+
+export class Charts extends OnInit {
+
+    private chartList: Array<String>;
+    private chartData: Object;
+
+    constructor(private chartService: ChartService){
+      super();
```
fungjj92: Explanation for doing this: http://stackoverflow.com/questions/31067368/javascript-es6-class-extend-without-super

src/ccl/charts/line-graph.component.ts
CloudNiner: ere's an example of doing it this way, I also have limited experience here so there might be a better way:
Define model:
https://gitlab.com/CloudNiner/disc.us/blob/feature/webapp/ng2/src/app/models/Discussion.ts
Retrieve data from server as instance of model:
https://gitlab.com/CloudNiner/disc.us/blob/feature/webapp/ng2/src/app/services/nwsoffice.service.ts#L19
Use data in component:
https://gitlab.com/CloudNiner/disc.us/blob/feature/webapp/ng2/src/app/office-detail/office-detail.component.ts

If you need to do a transformation in between pulling the data and providing to a component, that might be a bit more involved. Could probably have a function that returns the target type, and just creates an object and then typecasts it to start.

src/ccl/ccl.component.ts
```diff
@@ -3,20 +3,19 @@
...
- * App Component
- * Top Level Component
- */
 @Component({
   selector: 'ccl',
+  directives: [Sidebar, Charts],
   encapsulation: ViewEncapsulation.None,
   styleUrls: [
-  '../assets/css/main.css',
+    './ccl.style.css',
```
CloudNiner: The naming section of the angular style guide (https://angular.io/styleguide#!#02-01) suggests that this should be called `ccl.component.css` because its an asset for the ccl component.

src/ccl/ccl.component.ts
```diff
@@ -3,20 +3,19 @@
...
   ],
   templateUrl: './ccl.template.html'
```
CloudNiner: Same here, should be `ccl.component.html`. If there are other files named inconsistently in the project, they should be updated.

src/ccl/ccl.template.html
```diff
@@ -96,18 +60,7 @@ <h2 class="h4">Temperature</h2>
...
-                Total Precipitation
-              </h2>
-              <div class="chart-actions">
-                <button class="button button-link" data-toggle="tooltip" data-placement="bottom" data-delay="300" title="" data-original-title="Export"><i class="icon icon-export"></i></button>
-                <button class="button button-link" data-toggle="tooltip" data-placement="bottom" data-delay="300" title="" data-original-title="Options"><i class="icon icon-cog"></i></button>
-              </div>
-            </div>
-            <div class="chart-body"></div>
-          </div>
+          <charts></charts>
```
CloudNiner: Does everything (styling, etc) still work if you don't nest this under the div, and instead do `<charts class="charts scrollable"></charts>`? Would clean up the template a bit and remove an unnecessary div.

src/ccl/charts/charts-container.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    makeCharts() {
+      this.chartService.getChartData().subscribe(data=> {
+        this.chartData = data;
+      });
+    }
+
+    ngOnInit() {
+      // First, subscribe to chartList observable
+      this.chartList = this.chartService.get();
```
CloudNiner: Does an observable subscription need to be manually cleaned up in ngOnDestroy?

src/ccl/charts/charts-container.component.ts
CloudNiner: Hmm interesting. Haven't seen this before. Kind of annoying.

src/ccl/charts/charts-container.component.ts
```diff
@@ -0,0 +1,44 @@
...
+    './chart.style.css',
+  ],
+  template: `<div class="chart" *ngFor="let chart of chartList">
+                <chart [indicator]="chart" [chartData]="chartData"></chart>
+            </div>`
+})
+
+export class Charts extends OnInit {
+
+    private chartList: Array<String>;
```
CloudNiner: Here's another place where typescript could be advantageous. I would eventually expect this component to only require an array of Chart objects, returned from the chartService, rather than having to pull the chartList and the chartData separately.

Also ok with deferring this, but its worth thinking about because its still easy ATM to change the public interfaces of these components.

src/ccl/charts/line-graph.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  inputs: [ 'data', 'indicator' ],
+  styleUrls: [ './chart.style.css']
+})
+
+export class LineGraph {
+
+  public data: Object;
+  public indicator: String;
+
+  private host;        // D3 object referebcing host dom object
```
CloudNiner: You may want to provide types for these private variables.

CloudNiner:
It looks like there are a few components in a single `charts/` directory. To improve organization and to follow the style guide, each component should be in its own folder:
```
foo/foo.component.ts
foo/foo.component.scss
foo/foo.component.html
foo/foo.component.spec.ts
```

Since all of these components are related to drawing charts, you could nest them under a parent folder of some kind, but not necessary. I think we'll want to evaluate project structure and rearrange as necessary as the project grows.

CloudNiner:
LGTM after renaming the html/css files and splitting each component to its own folder. Consider the other comments, or add TODOs where appropriate.

Nice start!

src/ccl/charts/charts-container.component.ts
fungjj92: I looked into it and there are no destroy or unsubscribe properties on Observables...?? 
https://xgrommx.github.io/rx-book/content/observable/observable_instance_methods/index.html

src/ccl/charts/charts-container.component.ts
fungjj92: I also imagined chartList eventually being an array of objects, but it's not there yet. When data is able to be pulled in by indicator from the api then it will be time.

src/ccl/charts/line-graph.component.ts
fungjj92: they're all DOM or D3 very specific elements so I'm gonna say not possible

fungjj92:
Updated with typescript models and file renaming

src/ccl/charts/charts-container.component.ts
```diff
@@ -22,7 +23,7 @@ import { Chart } from './chart.component';
 export class Charts extends OnInit {
 
     private chartList: Array<String>;
-    private chartData: Object;
+    private chartData: Object<ChartData>;
```
CloudNiner: Can this just be `ChartData` instead of `Object<ChartData>` in all these places? I'm not sure using the object generic gains anything as every class inherits from object anyways.

CloudNiner:
+1!

PR #9, opened 2016-08-16
Feature/fix webpack asset build

CloudNiner:
Probably best to look at the commits individually, and ignore the first commit as it just adds pre-existing assets. Details about caveats in the second commit's log, pasted here for visibility:
```
Must not use styles or styleUrls property of
angular component, instead add new component scss
file in assets dir and include in main.scss.
See src/assets/sass/components/_chart.scss for example
```
I'm sure there's a way to support both global styles and component styles in a webpack build, but I don't really think its worth it to dig into now. If each component gets its own scoped sass file, a refactor would be trivial later.

cc-lab-icons

### Testing

This changeset adds additional dependencies and thus requires an `npm install` before bringing up the webpack server.


package.json
```diff
@@ -64,43 +55,40 @@
...
+    "tslint-loader": "^2.1.3",
     "typedoc": "^0.4.4",
     "typescript": "^2.0.0-beta",
-    "typings":"^1.0.4",
-
-    "awesome-typescript-loader": "1.1.1",
-    "angular2-template-loader": "^0.4.0",
-    "tslint-loader": "^2.1.3",
+    "typings": "^1.0.4",
     "url-loader": "^0.5.7",
```
fungjj92: url-loader no longer in use, can be removed

config/webpack.common.js
```diff
@@ -102,6 +102,7 @@ module.exports = {
       "window.Tether": "tether"
     })
   ],
+  postcss: () => [autoprefixer({browsers: ['last 2 versions']})],
```
fungjj92: what is the purpose of specifying the last 2 versions of major browsers? i couldn't find a good explanation in the docs.

fungjj92:
+1 otherwise
Single happy tear seeing the icons

config/webpack.common.js
CloudNiner: Eh, after reading through the docs again, there really isn't one. I assumed the default was 'latest version', but its not, its '> 1%, last 2 versions, Firefox ESR' which is more than enough for us ATM. I'll switch to default.

CloudNiner: Totally doesn't apply here, but I didn't know you could target your own sites usage data using browserlist. That's a potentially neat feature.

https://github.com/ai/browserslist#custom-usage-data

PR #12, opened 2016-08-23
Feature/city autocomplete

flibbertigibbet:
Adds autocompletion module and implements autocomplete for the city selection field. Depends on companion PR azavea/climate-change-api#60 for CORS support and full-text search on city name.

Also upgrades to angular 2 rc5, as the [autocomplete module](https://github.com/ng2-ui/ng2-auto-complete) added now uses rc5. The autocomplete module is in the source with light modifications, rather than added as an npm dependency. The modifications are to append the authorization header to the GET request for the API search, and to add a configurable option for formatting the selection display string.

Adds a `constants.ts` file for the API host and key; the example constants file will need to be copied and edited to run the project after this gets merged, as noted in the README changes.


src/main.browser.ts
```diff
@@ -6,17 +6,19 @@ import { bootstrap } from '@angular/platform-browser-dynamic';
 * Platform and Environment
```
fungjj92: The { bootstrap } import is no longer in use

src/main.browser.ts
```diff
@@ -6,17 +6,19 @@ import { bootstrap } from '@angular/platform-browser-dynamic';
...
 * our providers/directives/pipes
 */
+import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { PLATFORM_PROVIDERS } from './platform/browser';
 import { ENV_PROVIDERS, decorateComponentRef } from './platform/environment';
```
fungjj92: the *_PROVIDERS are not used anymore either

src/ccl/app.module.ts
```diff
@@ -0,0 +1,34 @@
...
+    BrowserModule,
+    FormsModule,
+    HttpModule,
+  ],
+  providers: [ // expose our Services and Providers into Angular's dependency injection
+    PLATFORM_PROVIDERS,
+    ENV_PROVIDERS
+  ]
+})
+export class ClimateChangeModule {
```
fungjj92: This file should be ccl.module.ts to stay consistent

src/ccl/.gitignore
```diff
@@ -0,0 +1 @@
+constants.ts
```
fungjj92: It makes sense to me to put constants.ts in the config folder

fungjj92: Ineffective comment: I'm not sure where I think constants.ts should live. Here is probs fine for now.

src/ccl/auto-complete/auto-complete.component.ts
```diff
@@ -0,0 +1,233 @@
...
+          (mousedown)="selectOne(item)"
+          [ngClass]="{selected: i === itemIndex}"
+          [innerHTML]="getFormattedList(item)"
+          ></li>
+    </ul>
+
+  </div>`,
+  providers: [ AutoComplete ],
+  styles: [`
+  @keyframes slideDown {
```
fungjj92: Styling like this is totally valid ng2 but since it's a big chunk let's put it in a sass file

flibbertigibbet: Considering this came directly from a copied project I think it makes sense to preserve it to be able to do diffs.

fungjj92: You're right esp because angular2 and the project are still under active development, as we are seeing..

src/ccl/ccl.component.ts
```diff
@@ -5,14 +5,44 @@
...
   name = 'Climate Lab';
+
+  public apiCities: string = apiHost + "city/?search=:keyword&format=json";
+  public cityModel;
+
+  // custom formatter to display list of options as City, State
+  public cityListFormatter(data: any): string {
+    let html: string = "";
+    html += data.properties.name ? `<span>${data.properties.name}, ${data.properties.admin}</span>`: data;
+    return html;
```
fungjj92: In this ternary, what was the reasoning behind defaulting to 'data'?  Similar question in two other spots

src/ccl/ccl.component.ts
```diff
@@ -5,14 +5,44 @@
...
+  // custom formatter to display string for selected city as City, State
+  public cityValueFormatter(data: any): string {
+      let displayValue: string = "";
+      displayValue += data && data.properties ? data.properties.name + ', ' + data.properties.admin : data.toString();
+      return displayValue;
+  }
+
+  // callback invoked when user picks a city
+  public onCitySelected(value: any): void {
+      console.log('selected city changed to:');
```
fungjj92: Maybe put a TODO: here

src/ccl/ccl.component.ts
flibbertigibbet: You mean for the variable name? That happens to be the variable name used by the default formatter in the autocomplete module.

src/ccl/ccl.component.ts
flibbertigibbet: Currently working on a separate branch to implement listening and chart refresh.

src/ccl/ccl.component.ts
flibbertigibbet: Oh, you mean for the default if the data properties are missing: also what the autocomplete module default formatter does. If the properties are missing it's bad data anyways, so I suppose having something is better than nothing for debugging purposes.

fungjj92:
Okie +1, nice deep dive into angular2 :)

package.json
```diff
@@ -28,23 +28,23 @@
...
-    "@angular/http": "2.0.0-rc.4",
-    "@angular/common": "2.0.0-rc.4",
-    "@angular/compiler": "2.0.0-rc.4",
-    "@angular/core": "2.0.0-rc.4",
-    "@angular/forms": "0.2.0",
-    "@angular/platform-browser": "2.0.0-rc.4",
-    "@angular/platform-browser-dynamic": "2.0.0-rc.4",
-    "@angular/platform-server": "2.0.0-rc.4",
-    "es6-shim": "^0.35.0",
+    "@angular/common": "^2.0.0-rc.5",
```
CloudNiner: These angular dependencies should probably be locked to rc5. It seems the http one is already, but the rest are not.

flibbertigibbet: Good call. Looks like the [webpack starter](https://github.com/AngularClass/angular2-webpack-starter/blob/master/package.json) project most of this is derived from is using rc5 for those.

Issue #5, opened 2016-08-11
Add authentication for API


CloudNiner:
Some options:
- Use API login, CORS, cookies
- Setup account in lab

flibbertigibbet:
azavea/climate-change-api#60 adds CORS support to server side

flibbertigibbet:
Angular 2 does not have interceptors, so there is not a straightforward way to add authentication credentials to all API requests as we have done in Angular 1 projects previously. See [this discussion](http://stackoverflow.com/questions/34464108/angular2-set-headers-for-every-request).

CloudNiner:
Interesting discussion. A wrapper http service for interacting with the climate API is probably the way to go here. Would be useful to break out to a separate library down the road for reusability.

CloudNiner:
Closed by #50 and #41

PR #14, opened 2016-08-24
Feature/query api

flibbertigibbet:
Query live API to load chart data instead of reading dummy data from local file. Also implements reloading data when the city selection changes.

Does considerable data transformation to match the format of the dummy data and defined model; would probably be better to come up with a new model and/or server response to minimize work for the browser.


flibbertigibbet:
Note that this adds a constant for the default city that will need to be added to your constants file.

flibbertigibbet:
Partially addresses #5 by adding server communication with predefined credentials.

fungjj92:
+1

PR #17, opened 2016-08-29
Feature/graph controls

fungjj92:
Thought processes:
- It ultimately made sense to me to pass variables rather than create observables. My understanding is that observables are good for handling more substantive data. Also, Input/Output passing allows utilization of ngOnChanges in line-graph-component.
- Chart controls are not in their own component because otherwise it would create unnecessary component communication complexity and bulk. Being in the chart component doesn't seem to do any damage and is a lighter weight option.
- Min/Max feeds data into line-graph component, but don't draw anything on the graph because min/max aren't implemented yet (check your console for print statements)
- I put coverage in gitignore but tbh I don't exactly know what it is

screen shot 2016-08-29 at 3 07 43 pm


src/ccl/charts/chart.component.html
```diff
@@ -1,13 +1,50 @@
...
+        [max]="max"
+        [minVal]="minVal"
+        [maxVal]="maxVal">
+    </line-graph>
 </div>
+<div class="chart-controls">
+    <h5>Options</h5>
+    <ui-switch (change)="toggleTrendline()" size="small"></ui-switch>Show Trendline
+    <ui-switch (change)="toggleMin()" size="small"></ui-switch>Show Min
+    <form>
```
CloudNiner: I haven't looked in the ngforms documentation much, but this feels odd. Creating two forms and using the standalone argument. Was there a technical reason for wrapping each of these inputs in a form element?

CloudNiner:
What's here looks good as implemented. I agree on the points above. 

It looks like this only adds the controls. Do we have a separate card for the settings cog toggle showing/hiding these controls? And to actually make the controls do what they say they do? Doesn't look like the min/max lines controls are hooked up.

fungjj92:
The controls work insofar as feeding the right signals info to the line-graph component. Min/Max is in slow d3 progress in my 'doing' box.

I missed the settings toggle, I'll add that and post an update when it's pushed.

src/ccl/charts/chart.component.html
fungjj92: Good question. The toggle buttons (3rd party lib) in between the number inputs are an issue if put into an ng-form.

CloudNiner: Does an ngModel now need to be nested within a form?

fungjj92: No

fungjj92: I tried using inputs (without ngForms) first but it didn't work as planned. IDR why though, but i can give it another go and find out

CloudNiner: Ah ok. Worth revisiting as our knowledge of this expands. Let's move forward with other tasks ATM though.

fungjj92: It worked! Quick fix to native input element. Whoot whoot.

fungjj92:
Updated! Note, I replaced bootstrap with ng2-bootstrap which has some useful custom directives in it. 
https://github.com/valor-software/ng2-bootstrap/tree/master

CloudNiner:
Looks like the files in `./coverage` got committed accidentally

src/vendor.browser.ts
```diff
@@ -13,7 +13,7 @@ import 'rxjs/add/operator/mergeMap';
 
 // 3rd Party Libraries
 import 'jquery';
-import 'bootstrap';
 import 'lodash';
 import 'd3';
 import 'd3-tip';
+import 'ng2-bootstrap';
```
CloudNiner: I don't think we're using the original bootstrap import for anything, as all of our styles are compiled from the bootstrap sass files in the assets directory. It may be better to just try removing the bootstrap import, rather than replacing it with ng2-bootstrap.

I'm also not convinced this is the proper way to use ng2-bootstrap. I expected it to be imported within individual components just like the UI switch component. Maybe I'm wrong though. Either way, if we're not currently using any of its components, it should just be removed until its needed.

src/ccl/ccl.module.ts
```diff
@@ -0,0 +1,36 @@
+import { NgModule } from '@angular/core';
```
CloudNiner: Looks like there was maybe a bad merge conflict resolution here. This file should no longer exist, any modifications you made to it should be moved to `app.module.ts`

src/vendor.browser.ts
fungjj92: Oh sorry I wasn't clear. ng2-boostrap houses the collapse feature I used.

CloudNiner: Oh shoot. Sorry, I missed the use of that component yesterday while scanning through. Using ng2-bootstrap sounds good, removing the bootstrap dependency is also good. 

Is this import here in vendor.browser.ts necessary? I would think the piecemeal imports in each of our source files would be enough.

fungjj92:
Bump :o)

CloudNiner:
+1 looks great

PR #18, opened 2016-08-29
Feature/routing

CloudNiner:
Adds basic routing to the project. Some remaining TODOs that can be addressed after this is merged (copied from commit log):
- ~~Rename existing components to match angular style guide~~
- Update top level project directory from ccl -> app
- Reorganize project so that components and services are
  better modularized
- Add PageNotFoundComponent

PR #19, opened 2016-08-30
Feature/routing part two

CloudNiner:
Part two of the routing task, which addresses the remaining items in #18 

I didn't feel like there was quite enough substance here to refactor to a bunch of submodules quite yet  -- I had trouble coming up with clean breaks for where certain things should go. We'll want to address this soon though as we start to add other pages, like the dashboard.


fungjj92:
Getting error "client:47 ./src/main.browser.ts
Module build failed: Error: Cannot resolve 'file' or 'directory' ../constants"

CloudNiner:
Did you ctrl+c and restart your webpack dev server after pulling this PR? Sometimes the dev server loses file handles when you move files around underneath it.

CloudNiner:
@fungjj92 bump. Did restarting the dev server make this go away? If not, did you `mv src/ccl/constants.ts src/app/`?

fungjj92:
ah duh we have a new root at /app not /ccl

fungjj92:
Thinking about the main template file looking sparse with just component tags is really exciting!!

Re: holding off on component-izing more--> The only other major part is the buttons bar, but that is actively being worked on and would just break stuff unnecessarily now. It may become a little complicated with the autocomplete but we'll cross that bridge.

LGTM

PR #21, opened 2016-08-31
Feature/select model

flibbertigibbet:
Implements two selectors in the navbar, one for changing the scenario, and the other for selecting models. There's not an obvious/easy way to switch back to selecting all models yet. If we move towards using ng2-angular components, I think it might be good to use their multi-select instead of the existing modal of checkboxes from the prototype that I wired up for the model selector here.

This moves the remaining default API query options into the `constants.ts` file, so there are three new constants that will need to be added to run this.

After rebasing off of develop, a number of new typechecker warnings appeared, mostly suggesting using `string` instead of `String`, so I have made those changes (there are no linter warnings now.)

### DIFF
--- a/example/constants.ts.example
+++ b/example/constants.ts.example
@@ -5,5 +5,9 @@
 
 
 export const defaultCity = {"id":7, "properties":{"name":"Philadelphia","admin":"PA"}};
+export const defaultScenario = 'RCP85';
+export const defaultVariable = 'pr';
+export const defaultYears = 2070;
+
 export const apiHost = "https://staging.api.futurefeelslike.com/api/";
 export const apiToken = "YOUR_API_TOKEN_HERE";
--- a/example/constants.ts.example
+++ b/example/constants.ts.example
@@ -7,7 +7,7 @@
 export const defaultCity = {"id":7, "properties":{"name":"Philadelphia","admin":"PA"}};
 export const defaultScenario = 'RCP85';
 export const defaultVariable = 'pr';
-export const defaultYears = 2070;
+export const defaultYears = '2070';
 
 export const apiHost = "https://staging.api.futurefeelslike.com/api/";
 export const apiToken = "YOUR_API_TOKEN_HERE";
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -17,7 +17,7 @@ import * as $ from 'jquery';
 export class LineGraphComponent {
   public data: ChartData[];
   public extractedData: Array<DataPoint>;
-  public indicator: String;
+  public indicator: string;
   public trendline: Boolean;
   public min: Boolean;
   public max: Boolean;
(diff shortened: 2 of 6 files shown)


CloudNiner:
@fungjj92 just merged her chart controls PR, which adds the ng2 bootstrap lib and removes the bootstrap lib. It looks like this PR makes use of the modal and dropdown bootstrap js, and needs a rebase to switch to the ng2 bootstrap versions of those components. 

Otherwise looks great!

flibbertigibbet:
There's no bootstrap JS in here, just styling classes and a TODO about maybe switching in future to an ng2-bootstrap component once it's stable.

CloudNiner:
There's JS somewhere that's making the modal/dropdowns work right now, I assumed it was the `import 'bootstrap'` in vendor.browser.ts, which jenny removed in her latest PR.

flibbertigibbet:
Okay, a rebase off of develop has broken the dropdowns (which were already present an functioning, in mock form). I'll look into it.

CloudNiner:
Yup, thats it. If i remove the `import bootstrap` line then the modals/dropdowns no longer work. So these controls should be updated to use the ng2-bootstrap versions

CloudNiner:
I updated my constants.ts file to include the new constants, does your webpack linter throw these errors:
```
[default] /home/andrew/src/climate-change-lab/src/app/lab/lab.component.ts:14:32 
    Module '"/home/andrew/src/climate-change-lab/src/app/constants"' has no exported member 'defaultScenario'.
[default] /home/andrew/src/climate-change-lab/src/app/services/chart.service.ts:7:42 
    Module '"/home/andrew/src/climate-change-lab/src/app/constants"' has no exported member 'defaultScenario'.
[default] /home/andrew/src/climate-change-lab/src/app/services/chart.service.ts:7:59 
    Module '"/home/andrew/src/climate-change-lab/src/app/constants"' has no exported member 'defaultVariable'.
[default] /home/andrew/src/climate-change-lab/src/app/services/chart.service.ts:7:76 
    Module '"/home/andrew/src/climate-change-lab/src/app/constants"' has no exported member 'defaultYears'.
```

flibbertigibbet:
The added constants were noted in the PR request.

flibbertigibbet:
No, I don't see those errors. It seems the linter requires a server restart to silence warnings related to changes, though it doesn't interfere with functionality.

flibbertigibbet:
The bootstrap library switch radically changes how components are used, as it requires loading directives now. The dropdown is a relatively straightforward fix, but the modal update is causing issues.

flibbertigibbet:
Ok, rebased off develop and fixed the dropdown and modal broken by the bootstrap library switch.

The backdrop on the modal has been disabled temporarily as a workaround for valor-software/ng2-bootstrap#854.

CloudNiner:
:+1:

PR #22, opened 2016-08-31
Feature/threshold lines

fungjj92:
## Overview

HOORAY. Draws user-set min and max threshold lines and fills in that space. Employs a bar graph with no spacing between bars. Styling aside, it is most beautiful and I am proud.

### Demo

screen shot 2016-08-31 at 5 27 16 pm

### Notes

I was having scope issues with "this" and d3 functions so I declared a "self" in ngOnInit for component-wide reference. IDK how legal this is as is but tslint seems to be happy.

## Testing Instructions

 Probably npm install just for good measure.

### DIFF
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -242,11 +242,10 @@ export class LineGraphComponent {
     }
   }
 
-  private drawLine(data: [], className: String): void {
+  private drawLine(data: Array<DataPoint>, className: String): void {
     this.svg.append("path")
       .data([data])
       .attr("class", className)
       .attr("d", this.valueline);
   }
-
 }
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -49,6 +49,8 @@ export class LineGraphComponent {
   /* Will Update on every @Input change */
   ngOnChanges(): void {
     if (!this.data || this.data.length === 0) return;
+    // Temporarily here to show chart control statuses
+    console.log("Min: " + this.min, this.minVal + "; Max: " + this.max, this.maxVal);
     this.filterData();
     this.setup();
     this.buildSVG();
@@ -57,8 +59,7 @@ export class LineGraphComponent {
     this.drawYAxis();
     this.populate();
     this.drawTrendLine();
-    // Temporarily here to show chart control statuses
-    console.log("Min: " + this.min, this.minVal + "; Max: " + this.max, this.maxVal);
+    this.drawMin();
   }
 
   private filterData(): void {
@@ -182,4 +183,44 @@ export class LineGraphComponent {
     return [slope, intercept, rSquare];
   }
 
+  private drawMin(): void {
+    var x1 = this.xRange[1];
+    var x2 = this.xRange[0];
+    var y = D3.max(this.yData)/2 // TODO: Accept user input as value
+    var minData = [{"date":x1, "value":y},{"date": x2, "value":y}];
+
+    this.svg.append("path")
+      .data([minData])
+      .attr("class", "minline")
+      .attr("d", this.valueline);
+
+    var xscale = this.xScale;
+    var yscale = this.yScale;
+
+    var areaBelowLine = D3.area()
+    .x(function(d) { return xscale(d.date); })
+    .y1(function(d) { return yscale(d.value); })
+    .y0(this.height);
+
+    var areaBelowMin = D3.area()
+    .x(function(d) { return xscale(d.date); })
+    .y(function(d) { return yscale(d.value); });
+
+    this.svg.append("path")
+      .attr("class", "area")
+      .data([minData])
+      .attr("d", areaBelowMin);
+
+    this.svg.append("path")
+      .attr("id", "clip-below")
+      .data([this.extractedData])
+      .attr("d", areaBelowLine.y0(this.height));
+
+    this.svg.append("path")
+      .attr("class", "areabelow")
+      .data([this.extractedData])
+      .attr("clip-path", "url(#clip-below)")
+      .attr("d", areaBelowLine);
+
+  }
 }
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -14,233 +14,223 @@ import * as $ from 'jquery';
   template: `<ng-content></ng-content>`,
   inputs: [ 'data', 'indicator', 'trendline', 'min', 'max', 'minVal', 'maxVal' ]
 })
+
 export class LineGraphComponent {
-  public data: ChartData[];
-  public extractedData: Array<DataPoint>;
-  public indicator: string;
-  public trendline: Boolean;
-  public min: Boolean;
-  public max: Boolean;
-  public minVal: number;
-  public maxVal: number;
 
-  private host;                          // D3 object referebcing host dom object
-  private svg;                           // SVG in which we will print our chart
-  private margin;                        // Space between the svg borders and the actual chart graphic
-  private width;                         // Component width
-  private height;                        // Component height
-  private xScale;                        // D3 scale in X
-  private yScale;                        // D3 scale in Y
-  private htmlElement;                   // Host HTMLElement
-  private valueline;                     // Base for a chart line
-  private xRange: Array<string>;         // Min, max date range
-  private xData: Array<number>;          // Stores x axis data as integers rather than dates, necessary for trendline math
-  private yData: Array<number>;          // Stores just y axis data, multi-use
-  private trendData: Array<DataPoint>;   // Formatted data for the trendline
+    public data: ChartData[];
+    public extractedData: Array<DataPoint>;
+    public indicator: string;
+    public trendline: Boolean;
+    public min: Boolean;
+    public max: Boolean;
+    public minVal: number;
+    public maxVal: number;
 
-  /* We request angular for the element reference
-  * and then we create a D3 Wrapper for our host element
-  */
-  constructor(private element: ElementRef) {
-    self = this;
-    self.htmlElement = self.element.nativeElement;
-    self.host = D3.select(self.element.nativeElement);
-  }
+    private host;                          // D3 object referebcing host dom object
+    private svg;                           // SVG in which we will print our chart
+    private margin;                        // Space between the svg borders and the actual chart graphic
+    private width;                         // Component width
+    private height;                        // Component height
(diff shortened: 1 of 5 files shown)


src/app/charts/line-graph.component.ts
```diff
@@ -42,121 +42,117 @@ export class LineGraphComponent {
   * and then we create a D3 Wrapper for our host element
   */
   constructor(private element: ElementRef) {
-    this.htmlElement = this.element.nativeElement;
-    this.host = D3.select(this.element.nativeElement);
+    self = this;
```
CloudNiner: This works, but very much not for the reason you expect. On this line, you don't declare it as `let self = this` which leads to the javascript default of `self` being declared as a global, which means that its accessible in this class, but also accessible _everywhere_. Which is not good, because it breaks encapsulation and clutters global scope. If you had declared it as `let self = this` then it would only be available in this function, and would be useless in all your other class methods.

Fortunately, there's a better way. And it's pretty nice in ES6. Anonymous arrow functions automatically bind to the `this` of the containing scope rather than the ES5 javascript default of binding to whatever the caller defines `this` to be, which defaults to `window`. So the somewhat hacky `var self = this` you would commonly see in older js code, such as:
```
function mapSomething() {
    var self = this;
    return _.map(this.data, function (d) { return self.someTransform(d); });
}
```
in ES6 can become the much cleaner:
```
function mapSomething() {
    return _.map(this.data, d => this.someTransform(d));
}
```

As always, the MDN docs are quite thorough, and worth read. More complete and detailed examples than the one provided above: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

src/app/charts/line-graph.component.ts
```diff
@@ -182,4 +178,69 @@ export class LineGraphComponent {
...
+
+        // Add bar graph
+        self.svg.selectAll(".min-bar")
+          .data(minBars)
+          .enter().append("rect")
+          .attr("class", "min-bar")
+          .attr("x", function(d) { return xscale(d.date); })
+          .attr("width", xscale.bandwidth())
+          .attr("y", function(d) { return self.yScale(d.value); })
+          .attr("height", function(d) { return self.height - self.yScale(d.value); });
```
CloudNiner: Here is an example line that would benefit from the comment above.

This line would become:
```
.attr("height", d => this.height - this.yScale(d.value));
```

Also in all other places in this diff where you use ES5 anonymous functions, e.g `function (d) { return d.foo; }` should be converted to use the ES6 arrow function syntax.

src/app/charts/line-graph.component.ts
```diff
@@ -182,4 +178,69 @@ export class LineGraphComponent {
...
+                    .domain(self.extractedData.map(function(d) { return d.date; }));
+
+      if (self.min && self.minVal) {
+        // Draw min threshold line
+        var minData = [{"date":x1, "value": self.minVal }, {"date": x2, "value": self.minVal }];
+        self.drawLine(minData, "minline");
+
+        // Prepare data for bar graph
+        var minBars = self.extractedData.map(function(datum) {
+            var val;
```
CloudNiner: Stylistically this is a bit convoluted. This would be much more concise for these two lines:
```
let val = this.minVal > datum["value"] ? y * 2 : 0;
```

src/app/charts/line-graph.component.ts
```diff
@@ -182,4 +178,69 @@ export class LineGraphComponent {
...
+            self.minVal > datum["value"]? val = y*2 : val = 0;
```
CloudNiner: Stylistically this is a bit convoluted. This would be far more concise, and properly uses the ternary operator:
```
let val = this.minVal > datum["value"] ? y * 2 : 0;
```

src/app/charts/line-graph.component.ts
```diff
@@ -182,4 +178,69 @@ export class LineGraphComponent {
     return [slope, intercept, rSquare];
   }
 
+  private drawMinMax(): void {
+    if (self.min || self.max) {
+      // Prepare standard variables
+      var x1 = self.xRange[1];
+      var x2 = self.xRange[0];
+      var y = D3.max(self.yData)/2;
```
CloudNiner: There are a few spots in this diff where spacing doesn't match our javascript style conventions. Strongly consider setting up something like eslint in your editor to warn you of these types of style violations.

src/app/charts/line-graph.component.ts
```diff
@@ -42,121 +42,117 @@ export class LineGraphComponent {
...
   private buildSVG(): void {
-    this.host.html('');
-    this.svg = this.host.append('svg')
-      .attr('width', this.width + this.margin.left + this.margin.right)
-      .attr('height', this.height + this.margin.top + this.margin.bottom)
-      .append('g')
-      .attr('transform', 'translate(' + this.margin.left + ',' + this.margin.top + ')');
+    self.host.html('');
+    self.svg = self.host.append('svg')
+                .attr('width', self.width + self.margin.left + self.margin.right)
```
CloudNiner: Inconsistent style here. These continuing lines should only be indented by two spaces

src/app/charts/line-graph.component.ts
```diff
@@ -42,121 +42,117 @@ export class LineGraphComponent {
...
     // Preserves parent data by fresh copying indicator data that will undergo processing
-    this.extractedData = _.cloneDeep(_.find(this.data, function(obj) {
+    self.extractedData = _.cloneDeep(_.find(self.data, function(obj) {
       return obj["indicator"] == indicator;
     }));
-    _.has(this.extractedData, "data")? this.extractedData=this.extractedData["data"] : this.extractedData=[];
+    _.has(self.extractedData, "data")? self.extractedData=self.extractedData["data"] : self.extractedData=[];
+    // Remove empty day in non-leap years
+    if (self.extractedData[365] && self.extractedData[365]['date'] == null) {
+        self.extractedData.pop(365);
```
CloudNiner: `pop()` takes no arguments

src/app/charts/line-graph.component.ts
```diff
@@ -182,4 +178,69 @@ export class LineGraphComponent {
...
+          .data(minBars)
+          .enter().append("rect")
+          .attr("class", "min-bar")
+          .attr("x", function(d) { return xscale(d.date); })
+          .attr("width", xscale.bandwidth())
+          .attr("y", function(d) { return self.yScale(d.value); })
+          .attr("height", function(d) { return self.height - self.yScale(d.value); });
+      }
+
+      if (self.max && self.maxVal) {
```
CloudNiner: Looks like you're copying almost the exact same code here for the max operation as above for the min operation. Can you think of a way to abstract this so that you only have to call a function twice rather than duplicate code?

src/app/charts/line-graph.component.ts
fungjj92: Ah! great this is what I would be looking for.

fungjj92:
Updated, and with 4 space indentation.

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+        this.drawYAxis();
+        this.populate();
+        this.drawTrendLine();
+        this.drawMinMax();
+    }
+
+    private filterData(): void {
+        let indicator = this.indicator;
+        // Preserves parent data by fresh copying indicator data that will undergo processing
+        this.extractedData = _.cloneDeep(_.find(this.data, obj => obj['indicator'] == indicator));
```
CloudNiner: ===

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+        _.has(this.extractedData, 'data')? this.extractedData=this.extractedData['data'] : this.extractedData=[];
```
CloudNiner: Spacing around = and ternary operator

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+        // Remove empty day in non-leap years
+        if (this.extractedData[365] && this.extractedData[365]['date'] == null) {
```
CloudNiner: ===

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+              // Calculate linear regression variables
+              var leastSquaresCoeff = this.leastSquares(this.xData, this.yData);
+
+              // Apply the results of the regression
+              var x1 = this.xRange[1];
+              var y1 = leastSquaresCoeff[0] + leastSquaresCoeff[1];
+              var x2 = this.xRange[0];
+              var y2 = leastSquaresCoeff[0] * this.xData.length + leastSquaresCoeff[1];
+              this.trendData = [{'date': x1, 'value': y1}, {'date': x2, 'value': y2}];
+              console.log(this.extractedData, this.trendData)
```
CloudNiner: Remove

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+            let y = D3.max(this.yData) / 2;
+
+            if (this.min && this.minVal) {
+                // Draw min threshold line
+                let minData = [{'date': x1, 'value': this.minVal}, {'date': x2, 'value': this.minVal}];
+                this.drawLine(minData, 'minline');
+
+                // Prepare data for bar graph
+                let minBars = _.map(this.extractedData, datum => {
+                    let val = this.minVal > datum['value']? y * 2 : 0;
```
CloudNiner: Spacing around ternary operator

src/app/charts/line-graph.component.ts
```diff
@@ -14,172 +14,224 @@ import * as $ from 'jquery';
...
+            }
+
+            if (this.max && this.maxVal) {
+                // Draw max threshold line
+                let maxData = [{'date': x1, 'value': this.maxVal}, {'date': x2, 'value': this.maxVal}];
+                this.drawLine(maxData, 'maxline');
+
+                // Prepare data for bar graph
+                let maxBars = _.map(this.extractedData, datum => {
+                    let val = this.maxVal < datum['value']? y * 2 : 0;
```
CloudNiner: Spacing

CloudNiner:
+1 after final spacing cleanup

CloudNiner:
Works great!

PR #24, opened 2016-08-31
Add option to select all models

### LINKED ISSUE #23: add option to reset model selection to 'all'
Modal with climate model checkboxes does not have an option to reset to default of all models, or a visually obvious way to show that is the default. Add a checkbox for the 'all' option that when checked, clears the other options.

Do not pass parameter with model list to API to get default of all models.

flibbertigibbet:
Add checkbox to top of model list for all models.
Checkbox is selected and disabled until at least one model gets picked.

Closes #23.

Also fixed chart tooltips and added descriptive tooltips to the scenario options.

### DIFF
--- a/src/app/charts/chart.component.html
+++ b/src/app/charts/chart.component.html
@@ -3,8 +3,12 @@
     {{indicator}}
     </h2>
     <div class="chart-actions">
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
image

CloudNiner:
:+1:

PR #34, opened 2016-09-02
Feature/indicator list from api

flibbertigibbet:
Retrieve list of indicators from API and use their descriptive labels in the sidebar.

Since the indicators do not have categories assigned in the API, the sidebar headers have been removed. Added tooltips with their descriptions on hover.

### DIFF
--- a/src/app/sidebar/sidebar.component.ts
+++ b/src/app/sidebar/sidebar.component.ts
@@ -2,6 +2,7 @@ import { Component, ViewEncapsulation, OnInit } from '@angular/core';
 
 import { ChartService } from '../services/chart.service';
 import { IndicatorsService } from '../services/indicators.service';
+import { Indicator } from '../models/indicator.models';
 
 /*
  * Sidebar Component
@@ -14,7 +15,7 @@ import { IndicatorsService } from '../services/indicators.service';
   templateUrl: './sidebar.component.html'
 })
 export class SidebarComponent extends OnInit {
-    private indicatorsList: string;
+    private indicators: Indicator[];
 
     constructor(private chartService: ChartService, private indicatorsService: IndicatorsService) {
       super();
@@ -22,12 +23,12 @@ export class SidebarComponent extends OnInit {
 
     // Set up click event handlers
     onIndicatorClicked(indicator) {
-        this.chartService.addChart(indicator);
+        this.chartService.addChart(indicator.name);
     }
 
     ngOnInit() {
-      this.indicatorsService.get()
-      .subscribe(data=>this.indicatorsList=data);
+      this.indicatorsService.loadIndicators();
+      this.indicatorsService.get().subscribe(data => this.indicators = data);
     }
 
 }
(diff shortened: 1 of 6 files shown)


src/app/sidebar/sidebar.component.ts
```diff
@@ -1,33 +1,38 @@
...
+    private indicators: Indicator[];
 
     constructor(private chartService: ChartService, private indicatorsService: IndicatorsService) {
       super();
     };
 
     // Set up click event handlers
     onIndicatorClicked(indicator) {
-        this.chartService.addChart(indicator);
+      // TODO: once indicator in place for raw data queries, change to pass Indicator object
```
fungjj92: If I'm following right I think the object doesn't need to be passed. I only see it being needed in the chartService. Please clarify if this is incorrect.

fungjj92:
It otherwise looks clean to me

fungjj92:
I am confused that the graphs are titled as "tasmin" when the indicator is yearly average * temperature

src/app/sidebar/sidebar.component.ts
flibbertigibbet: Right now it's querying the raw climate data endpoint. Once the indicator wrapper endpoint is in place we can change our queries to go to that endpoint and handle the charts in a consistent manner. We'd want to pass the full indicator object at that point to get its descriptive name, etc.

flibbertigibbet:
Graph title related to issue above; we're querying raw data still. Eventually we'll pass the indicator name to the endpoint instead of a variable name.

src/app/sidebar/sidebar.component.ts
CloudNiner: Is there a reason we can't switch to using the indicator endpoint right now? These graphs are supposed to show the indicators anyways, not the underlying data. Then you'd be able to pass the complete indicator object too.

flibbertigibbet: Our only functioning chart right now uses raw data (`pr`), and expects to do a bunch of transforms from the raw data format to the expected chart format.

CloudNiner: Ah, right. Is that something you envisioned being done as part of azavea/climate-change-api#68?

flibbertigibbet: Yes. Once a wrapper is in place for the raw data then we can just use the indicator endpoint and its format.

fungjj92:
Aside from comment +1 i think (fingers crossed for next step lol)

src/app/sidebar/sidebar.component.ts
```diff
@@ -1,33 +1,38 @@
...
+      this.chartService.addChart(indicator.variables[0]);
```
flibbertigibbet: Not if we want the current chart to continue to function. It's passing `pr` (the variable name) to the raw data endpoint. Point of comment is we can pass the whole indicator object later and use its label then.

PR #37, opened 2016-09-02
Feature/dashboard

fungjj92:
## Overview

Creates the dashboard page as the app homepage. Two view for if there are or aren't projects. Default set to false. The lab can now be found at [http://localhost:3100/#/lab](http://localhost:3100/#/lab) 

### Demo

screen shot 2016-09-02 at 4 34 51 pm

### Notes

Styling is extremely light. Regular bootstrap is broken...

## Testing Instructions

Change the this.isProject bool to true in dashboard.component.ts to see the alternate view.

### DIFF
--- a/src/app/dashboard/dashboard.component.html
+++ b/src/app/dashboard/dashboard.component.html
@@ -2,22 +2,29 @@
 
     <navbar></navbar>
 
-    <div class="dash">
-        <span class="dash-left">
+    <div class="dash-header">
+        <div class="dash-left">
             <h2>Lab</h2>
-        </span>
-        <span class="dash-right">
+        </div>
+        <div class="dash-right">
             <button type="button" class="btn btn-primary">New Project</button>
-        </span>
-        <div class="dash-center">
-            <!-- TODO: add ngIf flag for (yes/no) projects -->
+        </div>
+    </div>
+    <div class="dash-body">
+        <div *ngIf="!isProject">
             <i class="icon-cloud"></i>
-            <div>To explore climate change data, <a href="#/lab">start a new project</a></div>
-            <div>
+            <div>To explore climate change data, <a href="#/lab">start a new project.</a></div>
+        </div>
+        <div *ngIf="isProject">
+            <!-- TODO: Populate a table with eventual projects-->
         </div>
     </div>
     <div class="about">
-        <h2>API</h2>
+        <div class="title">
+            <h2>API</h2>
+        </div>
+        <div class="about-body">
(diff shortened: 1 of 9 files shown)


src/app/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,30 @@
...
+            <h2>Lab</h2>
+        </div>
+        <div class="dash-right">
+            <button type="button" class="btn btn-primary">New Project</button>
+        </div>
+    </div>
+    <div class="dash-body">
+        <div *ngIf="!isProject">
+            <i class="icon-cloud"></i>
+            <div>To explore climate change data, <a href="#/lab">start a new project.</a></div>
```
CloudNiner: `<a>` tags should use the `routerLink` directive

src/app/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,30 @@
+<div class="app-fullscreen">
+
+    <navbar></navbar>
+
+    <div class="dash-header">
+        <div class="dash-left">
+            <h2>Lab</h2>
+        </div>
+        <div class="dash-right">
+            <button type="button" class="btn btn-primary">New Project</button>
```
CloudNiner: This button should probably also trigger a change to the lab page

flibbertigibbet:
Looks like we'll need to pin the bootstrap package version now to avoid pulling in 1.1.1. Could you put that in this PR?

src/app/dashboard/dashboard.component.html
flibbertigibbet: Looks like we've got some custom classes defined in use elsewhere in the app; this should probably have the classes `button button-default` instead, to match what's on the 'reset' button on the charts page.

src/app/dashboard/dashboard.component.ts
```diff
@@ -0,0 +1,17 @@
...
+import { NavbarComponent } from '../navbar/navbar.component';
+
+@Component({
+  selector: 'dashboard',
+  templateUrl: './dashboard.component.html',
+  directives: [ NavbarComponent ]
+})
+export class DashboardComponent {
+
+    private isProject: boolean;
```
flibbertigibbet: Maybe rename to `haveProjects`? I'm guessing this is to track whether or not there are any projects to list. Or better yet, define the list of `projects` as an array for which we can check the length.

flibbertigibbet:
+1 from me after addressing above

PR #38, opened 2016-09-06
Feature/tavg charts

fungjj92:
## Overview

Pull graph data from indicators endpoint (only Annual T min and T max right now).

### Demo
screen shot 2016-09-06 at 10 01 17 am

### Notes

I equivocated about where it was sensible to query but settled back at the chart container. This is because re-querying (i.e changing model, scenario, city) will need to happen to all active graphs and I felt it was better organized and time-sensible (multiquery observable) even though we pass around more data. I left but commented out prior code to handle daily data because we will need it soon.

The one linter error is because URLSearchParams doesn't want an array even if it accepts them just fine.

## Testing Instructions

Nothing special. Note, if you select RCP45 stuff blanks out in some cases because there is no data.

### DIFF
--- a/src/app/charts/charts-container.component.ts
+++ b/src/app/charts/charts-container.component.ts
@@ -1,6 +1,5 @@
 import { Component, OnInit } from '@angular/core';
 
-import { ChartData } from '../models/chart.models';
 import { ChartService } from '../services/chart.service';
 import { ChartComponent } from './chart.component';
 
@@ -12,27 +11,18 @@ import { ChartComponent } from './chart.component';
   selector: 'charts',
   directives: [ChartComponent],
   template: `<div class="chart" *ngFor="let chart of chartList">
-                <chart [indicator]="chart" [chartData]="chartData"></chart>
+                <chart [indicator]="chart"></chart>
             </div>`
 })
 export class ChartsContainerComponent extends OnInit {
     private chartList: Array<String>;
-    private chartData: ChartData[];
 
     constructor(private chartService: ChartService) {
       super();
     }
 
-    makeCharts() {
-      this.chartService.loadChartData();
-      this.chartService.getChartData().subscribe(data=> {
-        this.chartData = data;
-      });
-    }
-
     ngOnInit() {
       // First, subscribe to chartList observable
       this.chartList = this.chartService.get();
-      this.makeCharts();
     }
 }
--- a/src/app/charts/charts-container.component.ts
+++ b/src/app/charts/charts-container.component.ts
@@ -11,18 +11,27 @@ import { ChartComponent } from './chart.component';
   selector: 'charts',
   directives: [ChartComponent],
   template: `<div class="chart" *ngFor="let chart of chartList">
-                <chart [indicator]="chart"></chart>
+                <chart [indicator]="chart" [chartData]="chartData"></chart>
             </div>`
 })
 export class ChartsContainerComponent extends OnInit {
     private chartList: Array<String>;
+    private chartData;
 
     constructor(private chartService: ChartService) {
       super();
     }
 
+    makeCharts() {
+      this.chartService.loadChartData();
+      this.chartService.getChartData().subscribe(data=> {
+        this.chartData = data;
+      });
+    }
+
     ngOnInit() {
       // First, subscribe to chartList observable
       this.chartList = this.chartService.get();
+      this.makeCharts();
     }
 }
--- a/src/app/charts/charts-container.component.ts
+++ b/src/app/charts/charts-container.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit } from '@angular/core';
 
+import { ChartData } from '../models/chart.models';
 import { ChartService } from '../services/chart.service';
 import { ChartComponent } from './chart.component';
 
@@ -16,7 +17,7 @@ import { ChartComponent } from './chart.component';
 })
 export class ChartsContainerComponent extends OnInit {
     private chartList: Array<String>;
-    private chartData;
+    private chartData: Array<ChartData>;
 
     constructor(private chartService: ChartService) {
       super();
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -102,7 +102,7 @@ export class LineGraphComponent {
         this.xRange = D3.extent(this.extractedData, d => d.date);
         this.xScale.domain(this.xRange);
         // Adjust y scale, prettify graph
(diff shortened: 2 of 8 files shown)


CloudNiner:
@flibbertigibbet can you take a look at this instead?

src/app/charts/line-graph.component.ts
```diff
@@ -96,20 +96,23 @@ export class LineGraphComponent {
 
     // Set axis scales
     private setAxisScales(): void {
-        var parseTime = D3.timeParse('%Y-%m-%d');
+        // Time scales only recognize annual and daily data
+        var parseTime = D3.timeParse('%Y'); //D3.timeParse('%Y-%m-%d');
```
flibbertigibbet: This will only parse yearly aggregations. I suppose we should have some way to know or detect the aggregation level for a given chart. Not sure if we should rely on the client knowing that or not; maybe we could add the date aggregation level to the indicator listing API response and use that.

src/app/charts/charts-container.component.ts
```diff
@@ -17,7 +17,7 @@ import { ChartComponent } from './chart.component';
 })
 export class ChartsContainerComponent extends OnInit {
     private chartList: Array<String>;
-    private chartData: ChartData[];
+    private chartData: Array<ChartData>;
```
flibbertigibbet: There are a number of other places in the code defining things as `Thing[]`. Neither declaration form seem to displease the linter. Is there a TypeScript preference for the `Array<Thing>` syntax?

src/app/charts/line-graph.component.ts
```diff
@@ -96,20 +96,23 @@ export class LineGraphComponent {
...
         this.extractedData.forEach(d => d.date = parseTime(d.date));
         this.xRange = D3.extent(this.extractedData, d => d.date);
         this.xScale.domain(this.xRange);
-        this.yScale.domain([0, D3.max(this.yData)]);
+        // Adjust y scale, prettify graph
+        let yPad = Math.abs((D3.max(this.yData) - D3.min(this.yData)) * 2/3);
```
flibbertigibbet: Numeric constants generally should be declared with a `const` assignment to ease reading and potential reuse.

src/app/charts/line-graph.component.ts
fungjj92: Yea I thought about a flag but didn't yet because we only have yearly data
It would be ideal if it came through on the indicator API response

src/app/charts/line-graph.component.ts
```diff
@@ -96,20 +96,23 @@ export class LineGraphComponent {
...
 
     /* Will draw the X Axis */
     private drawXAxis(): void {
         this.svg.append('g')
           .attr('transform', 'translate(0,' + this.height + ')')
           .call(D3.axisBottom(this.xScale)
-          .ticks(5)
-          .tickFormat(D3.timeFormat('%m-%Y')));
+          .ticks(3)
+          .tickFormat(D3.timeFormat('%Y')));
```
flibbertigibbet: Related to other date formatting comment: the formatting string for a given chart should probably go in a chart-dependent variable, so it can be both reused within a given chart and also modified for different aggregation levels across charts.

src/app/charts/charts-container.component.ts
fungjj92: Both are equally accepted. I changed this to see if it would fix linter complaints and it didn't. I just didn't both to change it back.

src/app/charts/line-graph.component.ts
```diff
@@ -131,22 +134,19 @@ export class LineGraphComponent {
     private drawTrendLine(): void {
         // Only draw if data and add trendline flag
         if (this.trendline && this.extractedData.length) {
-          // Set up if first time
-          if (!this.trendData) {
```
flibbertigibbet: Is there no need to check for `trendData`? (Unsure if the check might have been lost with the indentation change.)

fungjj92:
Klaas's changes (adding indicators) has broken this. Will update when fixed.

fungjj92:
Also there are server errors 👎

src/app/services/chart.service.ts
```diff
@@ -147,6 +152,29 @@ export class ChartService {
...
+            if (!_.includes(indicators, indicator)) {
+                indicators.push(indicator);
+                chartData.push({
+                    'indicator': indicator,
+                    'data': indicatorData
+                } as ChartData);
+            }
+        });
+
+        /* For daily value conversions
```
flibbertigibbet: Remove.

(If we need this again, it's in the commit history.)

src/app/charts/charts-container.component.ts
flibbertigibbet: Okay. No biggie, but I suppose we should pick one as a team style then.

src/app/charts/line-graph.component.ts
fungjj92: I'd put a check for trendData before to prevent unnecessary recalculations. Since data is no longer static, we'll want to recalculate lines.

flibbertigibbet:
:+1: 

I noticed that on window resize the charts do not redraw; I will make an issue for that.

fungjj92:
Updated :)

src/app/charts/line-graph.component.ts
```diff
@@ -97,12 +106,12 @@ export class LineGraphComponent {
...
     private setAxisScales(): void {
         // Time scales only recognize annual and daily data
-        var parseTime = D3.timeParse('%Y'); //D3.timeParse('%Y-%m-%d');
+        var parseTime = D3.timeParse(this.timeFormat);
         this.extractedData.forEach(d => d.date = parseTime(d.date));
         this.xRange = D3.extent(this.extractedData, d => d.date);
         this.xScale.domain(this.xRange);
         // Adjust y scale, prettify graph
-        let yPad = Math.abs((D3.max(this.yData) - D3.min(this.yData)) * 2/3);
+        const yPad = ((D3.max(this.yData) - D3.min(this.yData)) > 0) ? ((D3.max(this.yData) - D3.min(this.yData)) * 2/3) : 5;
```
flibbertigibbet: Actually meant to suggest 2/3 be made a constant, but I suppose this works too. Generally numbers in code should be assigned to constants for readability. (What does 2/3 do here? Is that a spacing amount?)

flibbertigibbet:
:+1:

src/app/charts/line-graph.component.ts
fungjj92: Oh yeah, it is y axis padding on either side of the data so that the graph doesn't hug the data exactly.

PR #41, opened 2016-09-06
Add Http wrapper

### LINKED ISSUE #27: HTTP service wrapper for auth

flibbertigibbet:
Closes #27.
Append auth header to requests to API.
Also check if request fails as forbidden, and if so, redirect to login.

### DIFF
--- a/src/app/auth/api.http.ts
+++ b/src/app/auth/api.http.ts
@@ -44,7 +44,7 @@ export class ApiHttp extends Http {
         if (options.headers == null) {
             options.headers = new Headers();
         }
-        options.headers.append('Authorization', 'Token ' + apiToken);
+        options.headers.set('Authorization', 'Token ' + apiToken);
         return options;
     }
 
--- a/src/app/auth/api.http.ts
+++ b/src/app/auth/api.http.ts
@@ -1,0 +1,70 @@
+import { Injectable } from '@angular/core';
+import { ConnectionBackend, Http, Headers, Request, RequestOptions, RequestOptionsArgs, Response } from '@angular/http';
+import { Observable } from "rxjs";
+
+import { apiToken } from "../constants";
+
+import * as _ from 'lodash';
+
+/**
+ * Wrapper for Http that appends authorization headers for requests to the API backend,
+ * and redirects unauthorized responses to login page.
+ */
+@Injectable()
+export class ApiHttp extends Http {
+
+    constructor(protected _backend: ConnectionBackend, protected _defaultOptions: RequestOptions) {
+        super(_backend, _defaultOptions);
+    }
+
+    request(url: string | Request, options?: RequestOptionsArgs): Observable<Response> {
+        return this.intercept(super.request(url, this.appendAuthHeader(options)));
+    }
+
+    get(url: string, options?: RequestOptionsArgs): Observable<Response> {
+        return this.intercept(super.get(url, this.appendAuthHeader(options)));
+    }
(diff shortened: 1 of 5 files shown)


flibbertigibbet:
Loosely based on [this](https://www.illucit.com/blog/2016/03/angular2-http-authentication-interceptor/).

src/app/auth/api.http.ts
```diff
@@ -0,0 +1,70 @@
...
+import { apiToken } from "../constants";
+
+import * as _ from 'lodash';
+
+/**
+ * Wrapper for Http that appends authorization headers for requests to the API backend,
+ * and redirects unauthorized responses to login page.
+ */
+@Injectable()
+export class ApiHttp extends Http {
```
CloudNiner: This class feels more generic than being an Http class for an API. More like `TokenAuthHttp` maybe? Since all it really does is append a token authorization header.

src/app/auth/api.http.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    appendAuthHeader(options?: RequestOptionsArgs) : RequestOptionsArgs {
+        if (options == null) {
+            options = new RequestOptions();
+        }
+        if (options.headers == null) {
+            options.headers = new Headers();
+        }
+        options.headers.set('Authorization', 'Token ' + apiToken);
```
CloudNiner: Rather than import constants directly here, could we inject it at app bootstrap using providers? This would allow us to reuse this class in other projects.

See: https://angular.io/docs/ts/latest/guide/dependency-injection.html#!#dependency-injection-tokens

src/app/auth/api.http.ts
```diff
@@ -0,0 +1,70 @@
...
+        return options;
+    }
+
+    /**
+     * Check if request forbidden, and if so, redirect to login page.
+     */
+    intercept(observable: Observable<Response>): Observable<Response> {
```
CloudNiner: Nice! I'll use this for the login component task.

CloudNiner:
+1 consider comments

src/app/auth/api.http.ts
flibbertigibbet: Well, it appends the token specifically for our climate data API.

src/app/auth/api.http.ts
flibbertigibbet: Not sure how that will work with the weird constructor/factory stuff in `browser-providers.ts`, but I'll give it a shot. That would allow us to also re-use this with other APIs in the same project too, I guess.

src/app/auth/api.http.ts
CloudNiner: True, although with change below it could be used for any requests that require token auth, not just our API.

src/app/auth/api.http.ts
CloudNiner: Yeah, that would definitely work as well!

flibbertigibbet:
Attempting to inject an `OpaqueToken` as suggested in the docs is failing for reasons that are... opaque. The `ApiHttp` class appears to instantiate fine, but then is not available to the services where it should be injected; the actual error message happens when a `get` attempt happens, but the `ApiHttp` object on the service is missing. I'll merge this as-is, and if we get some more clues on how to properly handle chaining DI we can revisit.

CloudNiner:
Bummer. Could have to do with not being on the latest RC. Thanks for taking a look!

PR #46, opened 2016-09-07
Resize graphs on window width change

### LINKED ISSUE #40: Redraw charts on window resize
Otherwise chart lines overflow horizontal space.

fungjj92:
Fixes #40 

The listener is set up in the charts container so that we only have 1 listener. It seemed a cleaner idea than attaching a listener to every chart component or graph even if it means passing around another input variable another layer.

### DIFF
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -1,4 +1,4 @@
-import { Component, ViewEncapsulation, ElementRef } from '@angular/core';
+import { Component, ViewEncapsulation, ElementRef, HostListener } from '@angular/core';
 import { ChartData, DataPoint } from '../models/chart.models';
 import { Indicator } from '../models/indicator.models';
 import * as D3 from 'd3';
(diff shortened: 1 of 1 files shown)


CloudNiner:
I'm not so sure. By setting it up this way, now you require that a line-graph component be within a chart component which must be within a charts-container component in order for it to work correctly, which breaks the encapsulation of line-graph being self-sufficient. 

Since the line-graph is the one that cares directly about the resize element, the listener should probably go there. That way the line-graph component can go anywhere, and still properly handle its own drawing without dependencies.

fungjj92:
Hm I see your point

fungjj92:
I updated this. I haven't seen any documentation about best practices. It feels weird to manually call ngOnChanges even if it works :/

src/app/charts/line-graph.component.ts
```diff
@@ -55,6 +55,11 @@ export class LineGraphComponent {
         }
     }
 
+    @HostListener('window:resize', ['$event'])
+    onResize(event) {
+      this.ngOnChanges();
```
CloudNiner: This seems ok as long as we don't ever need the changed values from one of the `@Input()` properties that get passed to ngOnChanges. 

As an aside, since in the resize case, presumably the data isn't changing, and the chart has already been drawn, could we only perform a subset of the calculations in ngOnChanges to re-render?

CloudNiner:
+1

PR #50, opened 2016-09-07
Feature/login component

CloudNiner:
## Overview

Implements the login component, and an AuthService to handle login/logout as well as storage and retrieval of the API token.

### Demo

cclab-login-2

### Notes

TODO:
- [x] Cleanup styling of login page
- [x] Store token via LocalStorage so that it persists on refreshes and across tabs 

## Testing Instructions

 * Update constants.ts by removing apiToken and updating apiHost to only include the hostname component of the URL
 * Navigate to [http://localhost:3100]() - The app should redirect to /login
 * Login using your username/password from the staging API
 * Browse the app as before
 * Refreshing the page should keep you logged in

### DIFF
--- a/src/app/auth/auth.service.ts
+++ b/src/app/auth/auth.service.ts
@@ -33,6 +33,10 @@ export class AuthService {
         }
     }
 
+    getToken(): string {
+        return this.token || '';
+    }
+
     isAuthenticated() {
         return !!this.token;
     }
--- a/src/app/auth/auth.service.ts
+++ b/src/app/auth/auth.service.ts
@@ -1,0 +1,39 @@
+import { Injectable } from '@angular/core';
+import { Headers, Http, RequestOptions, Response } from '@angular/http';
+import { Router } from '@angular/router';
+import { Observable } from "rxjs";
+
+import { apiHost } from '../constants';
+
+@Injectable()
+export class AuthService {
+
+    private token: string;
+
+    constructor(protected http: Http, protected router: Router) {}
+
+    login(username: string, password: string): Observable<any> {
+        let body = JSON.stringify({
+            'username': username,
+            'password': password
+        });
+        let headers = new Headers({ 'Content-Type': 'application/json' });
+        let options = new RequestOptions({ headers: headers });
+        let url = `${apiHost}/api-token-auth/`;
+        return this.http.post(url, body, options).map(response => {
+            let token = response.json().token;
+            this.token = token;
+        });
+    }
+
+    logout(redirectTo: string = '/login') {
+        this.token = null;
+        if (redirectTo) {
+            this.router.navigate([redirectTo]);
+        }
+    }
+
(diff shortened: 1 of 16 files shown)


src/app/auth/auth.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    login(username: string, password: string): Observable<any> {
+        let body = JSON.stringify({
+            'username': username,
+            'password': password
+        });
+        let headers = new Headers({ 'Content-Type': 'application/json' });
+        let options = new RequestOptions({ headers: headers });
+        let url = `${apiHost}/api-token-auth/`;
+        return this.http.post(url, body, options).map(response => {
+            let token = response.json().token;
```
flibbertigibbet: Shouldn't we do some kind of error handling here to handle bad login credentials?

src/app/auth/auth.service.ts
```diff
@@ -0,0 +1,49 @@
...
+            this.setToken(token);
+        });
+    }
+
+    logout(redirectTo: string = '/login') {
+        this.setToken(null);
+        if (redirectTo) {
```
flibbertigibbet: Maybe I'm not reading this correctly, but if there's a default value for the parameter, shouldn't this always be true?

src/app/login/login.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  templateUrl: './login.component.html'
+})
+export class LoginComponent {
+
+    constructor(private router: Router, private authService: AuthService) {}
+
+    createAccountUrl: string = `${apiHost}/accounts/register/`
+    resetPasswordUrl: string = `${apiHost}/accounts/password/change/`
+
+    model = new LoginForm('', '');
```
flibbertigibbet: Yay forms!

flibbertigibbet:
Autocomplete `apiHost` reference needs to be updated as well.

src/app/auth/auth.service.ts
CloudNiner: Yes, if no parameter is provided logout. You could pass a falsy value to logout to prevent a redirection.

src/app/auth/auth.service.ts
CloudNiner: This returns an observable, in our case the login component handles errors via `subscribe()`. I just pushed a commit to better handle failed logins by clearing the form.

flibbertigibbet:
If `canActivate` in `AuthGuard` is handling redirects for unauthenticated requests, then the `intercept` method in `ApiHttp` should be removed.

CloudNiner:
Ah, yes. Will remove.

Related, doesn't look like the delete method appends the auth header. I feel like it should?

flibbertigibbet:
Yes, `delete` should also use the auth header wrapper.

flibbertigibbet:
:+1:

Issue #36, opened 2016-09-02
Update ng2-bootstrap version


CloudNiner:
Based on this comment, v1.1.0 supports rc5 and v1.1.1+ supports rc6

https://github.com/valor-software/ng2-bootstrap/issues/854#issuecomment-244386904

flibbertigibbet:
Looks like a fix just got merged for the modal background issue: valor-software/ng2-bootstrap/issues/854

PR #58, opened 2016-09-12
Feature/angular rc6

### LINKED ISSUE #35: Upgrade to ng2-rc6

CloudNiner:
## Overview

Updates the Climate Lab app to Angular RC6. Being on older versions limits our ability to use up to date versions of third party libraries, and causes mismatches between our code and samples/howtos in the Angular official documentation.

### Notes

- ng2-bootstrap updated to v1.1.1
- Pulled in changes to ng2-auto-complete plugin, equivalent to v0.5.5, and applied an additional bugfix for RC6
- Bugfixed a change in the Indicator API endpoint transformer due to azavea/climate-change-api#91
- Bugfixed an issue with our auto-complete customization where the valueFormatter would only apply if the dropdown was open
- Moved auto-complete module to folder name matching the node package name for consistency

## Testing Instructions

* A clean install of node modules is best, so `rm -rf node_modules && npm install` before bringing up the dev server with `npm start`
* App should continue to function as before

Closes #35, #36


src/app/ng2-auto-complete/ng2AutoComplete.module.ts
```diff
@@ -0,0 +1,28 @@
...
+    provide: ApiHttp,
+    useFactory: (xhrBackend: XHRBackend, requestOptions: RequestOptions, authService: AuthService) => {
+        return new ApiHttp(xhrBackend, requestOptions, authService)
+    },
+    deps: [XHRBackend, RequestOptions, AuthService]
+};
+
+@NgModule({
+  imports: [ BrowserModule, FormsModule ],
+  entryComponents: [AutoCompleteComponent],
```
CloudNiner: An error after upgrading stumped me for a bit, where the new RC6 `ComponentFactoryResolver` couldn't find a component with the name of `AutocompleteComponent` even though it was properly defined and imported. Then I found some doc notes by browsing the RC6 changelog and related issues noting that any components dynamically created via the `ComponentFactoryResolver` need to be declared in this `entryComponents` property.

flibbertigibbet:
It looks like the latest release for the autocomplete module is 0.6.2. Why use 0.5.5?

Regarding the bugfix for the `valueFormatter`, is that something new that appeared? Maybe [this issue](https://github.com/ng2-ui/ng2-auto-complete/issues/14)? Had not seen that before, and unclear why that would be specific to our customization.

src/app/ng2-auto-complete/README.md
```diff
@@ -3,4 +3,3 @@
 This module is a lightly modified version of [ng2-auto-complete](https://github.com/ng2-ui/ng2-auto-complete).
```
flibbertigibbet: Might be nice to also note the release version we're using here.

CloudNiner:
Because 0.5.5 was the latest release on Sat morning when I was working on this. The 0.6 releases were pushed over the weekend.

flibbertigibbet:
I take it you mean Friday morning. Okay.

flibbertigibbet:
:+1:

CloudNiner:
Re the valueFormatter: 

I'm not sure if the valueFormatter issue is new, it may have been masked for various reasons that were exposed in the RC upgrade. That issue is not related, and none of the other changes released this weekend appear to be either.

The problem with that is that the AutocompleteComponent only exists when the dropdown is open. We were manually setting the `cityModel` property in the LabComponent onInit hook, which is a time when the dropdown, and thus its valueFormatter, doesn't exist. This led to 'undefined' showing in the input box rather than the proper city string. In addition, we were setting `cityModel` to a string, and it appears that it should be a City object. So this fixes that up and ensures that the valueFormatter always exists, and that the cityModel is always of the proper type.

CloudNiner:
Whoops. Yes I do.

src/app/ng2-auto-complete/README.md
CloudNiner: Will do.

flibbertigibbet:
Okay. That issue must be related to the upgrade then. The initial city value had to be set to a string previously, or it would display `undefined`, as it expected the formatted value at that point and not the model. Not sure how the dropdown could be open if it doesn't exist yet.

Seems to function fine with these changes.

CloudNiner:
Ah. Yup, the behavior of the auto-complete module has changed.

Issue #52, opened 2016-09-08
Deploy app via jenkins


CloudNiner:
Setup a s3 bucket to host

hectcastro:
It would be great to have the AWS resources captured in Terraform. Probably not a perfect fit, but [azavea-data-hub](https://github.com/azavea/azavea-data-hub/tree/master/deployment/terraform) has a Terraform 0.7 compatible setup for CloudFront with a S3 bucket origin. May need to tweak it so that it is a S3 website origin instead.

PR #59, opened 2016-09-12
Add missing clean:dist npm script

CloudNiner:
## Overview

This is the only _required_ change to get a working dist build. There are still issues where I can't seem to get files in `src/app/assets` to copy properly into the dist build. But that only affects display, not functionality. This unblocks #52.

## Testing Instructions

 * `npm run build:dist`
 * Start an http-server of your choice on a port of your choice, serving the files generated in the `./dist/` folder
 * The app should work with only the static image assets missing (which for now is only the svg image in the navbar


package.json
```diff
@@ -15,6 +15,7 @@
     "webpack-dev-server": "webpack-dev-server",
     "protractor": "protractor",
     "build": "npm run build:dev",
+    "clean:dist": "rm -rf dist",
```
flibbertigibbet: We may only have need to run this on *nix systems now, but it would generally be better to use some platform-independent build tool like [clean-webpack-plugin](https://github.com/johnagan/clean-webpack-plugin) for this task.

flibbertigibbet:
The build command seems to actually be `npm run build:prod`, though the clean task is `clean:dist`, so might consider renaming one or the other.

flibbertigibbet:
Tested via Web Server for Chrome.

:+1:

package.json
CloudNiner: True. I'll see about switching out for the plugin.

CloudNiner:
Yeah, its more generic to use `clean:prod`. I'll switch to that.

PR #60, opened 2016-09-13
Feature/min max

fungjj92:
## Overview

Adds a min.
Lab continuation to this workflow, CC-API PR: https://github.com/azavea/climate-change-api/pull/85

### Demo
screen shot 2016-09-13 at 2 58 59 pm

### Notes

Ideally we would draw 1 area that is a clipping of the two but I couldn't figure out how to do it. Instead, I drew two area bands: above min and above max. Setting the max area color to white makes sure only the desired area is shown. So it looks like a clip even if it's not truly.

## Testing Instructions

None.


src/app/charts/line-graph.component.ts
```diff
@@ -108,50 +113,50 @@ export class LineGraphComponent {
...
+          .y(d => this.yScale(d.value));
     }
 
     /* Will draw the X Axis */
     private drawXAxis(): void {
         this.svg.append('g')
           .attr('transform', 'translate(0,' + this.height + ')')
           .call(D3.axisBottom(this.xScale)
-          .ticks(3)
+          .ticks(5)
```
KlaasH: Worth defining these `ticks` arguments as constants at the top?

src/app/charts/line-graph.component.ts
```diff
@@ -186,21 +191,40 @@ export class LineGraphComponent {
...
+          .data([this.prepareLineData(this.yMinData)])
+          .attr('class', "area")
+          .attr('d', area);
+
+        // Draw area above maxline (not ideal, but visually clips the band)
+        this.svg.append('path')
+          .data([this.prepareLineData(this.yMaxData)])
+          .attr('class', "above-area")
+          .attr('d', area);
+    }
```
fungjj92: MaD3gician!

src/app/charts/line-graph.component.ts
```diff
@@ -86,7 +89,9 @@ export class LineGraphComponent {
             this.extractedData.pop();
         }
         // Parse out data by axis for ease of use later
-        this.yData = _.map(this.extractedData, d => d.value);
+        this.yAvgData = _.map(this.extractedData, d => d.values.avg);
+        this.yMinData = _.map(this.extractedData, d => d.values.min);
+        this.yMaxData = _.map(this.extractedData, d => d.values.max);
```
fungjj92: Mm you're right that not all of these are actually useful.

src/app/charts/line-graph.component.ts
```diff
@@ -196,11 +219,12 @@ export class LineGraphComponent {
...
                 // Draw min threshold line
                 let minData = [{'date': x1, 'value': this.minVal}, {'date': x2, 'value': this.minVal}];
-                this.drawLine(minData, 'minline');
+                this.drawLine(minData, 'min-threshold');
 
                 // Prepare data for bar graph
+                let yDataCopy = _.cloneDeep(this.yData);
                 let minBars = _.map(this.extractedData, datum => {
-                    let val = this.minVal > datum['value'] ? y * 2 : 0;
+                    let val = this.minVal >  yDataCopy.shift() ? y * 2 : 0;
```
KlaasH: I continue to find this shift-from-a-copy construction odd, though it does work fine.

One possible alternative, obviously, would just be using `datum.values.avg`.  If it's desirable to use `yData` (which I could see, to avoid adding two more references to the specific structure of the response data), another option would be to use the index provided by `_.map`, i.e.
```
let minBars = _.map(this.extractedData, (datum, index) => {
    let val = this.minVal >  this.yData[index] ? y * 2 : 0;
    ...
```

PR #61, opened 2016-09-13
Feature/vagrant build

flibbertigibbet:
Add configuration to optionally build project within vagrant, to be used for testing on Jenkins.

`vagrant up` brings up a VM with the project mounted at `/opt/climate-change-lab` and runs `npm install`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,86 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.require_version ">= 1.6"
+
+# Deserialize Ansible Galaxy installation metadata for a role
+def galaxy_install_info(role_name)
+  role_path = File.join("deployment", "ansible", "roles", role_name)
+  galaxy_install_info = File.join(role_path, "meta", ".galaxy_install_info")
+
+  if (File.directory?(role_path) || File.symlink?(role_path)) && File.exists?(galaxy_install_info)
+    YAML.load_file(galaxy_install_info)
+  else
+    { install_date: "", version: "0.0.0" }
+  end
+end
+
+# Uses the contents of roles.yml to ensure that ansible-galaxy is run
+# if any dependencies are missing
+def install_dependent_roles
+  ansible_roles_spec = File.join("ansible", "roles.yml")
+
+  YAML.load_file(ansible_roles_spec).each do |role|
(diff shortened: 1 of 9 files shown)


Vagrantfile
```diff
@@ -0,0 +1,86 @@
...
+                end
+
+ROOT_VM_DIR = "/opt/climate-change-lab"
+
+Vagrant.configure("2") do |config|
+
+  config.vm.define "app" do |app|
+    app.vm.box = "ubuntu/precise64"
+    app.vm.hostname = "app"
+    app.vm.network "private_network", ip: ENV.fetch("CLIMATE_CHANGE_PRIVATE_IP", "192.168.8.111")
```
CloudNiner: We should probably add a port forwarding entry here.

CloudNiner:
It would be great to have additional documentation/instructions in the 'Setup' section of the README to note this vm as an additional option for folks who don't want to install node on their host machine.

CloudNiner:
On my mac, after loading up the vm and doing `npm start` from within it, I got this error:
```
ERROR in Missing binding /opt/climate-change-lab/node_modules/node-sass/vendor/linux-x64-48/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 6.x

Found bindings for the following environments:
  - OS X 64-bit with Node.js 6.x

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to build the binding for your current environment.
 @ ./src/assets/sass/main.scss 4:14-175 13:2-17:4 14:20-181
```
Doing as the error suggests and running `npm rebuild node-sass` as suggested fixed things up. Could add that command to the provisioner since it wouldn't hurt to run it even if it doesn't need to, and it would save people who are switching OS environments an extra step.

flibbertigibbet:
Didn't see the `node-sass` issue, as I'm on a Linux host. Does that happen when setting up the project for the first time outside of vagrant? If so, that could go in the README.

I'll make the other changes for development usage (was setting it up just for jenkins.)

CloudNiner:
Yeah, I don't think you would see the node-sass issue. It happens if you `npm install` on the host and then `npm install` in the vm, or vice-versa, only if the host and vm are different OSes.

CloudNiner:
+1

PR #62, opened 2016-09-14
Feature/webpack prod build

### LINKED ISSUE #53: Test webpack production build

CloudNiner:
## Overview

This changset makes the last few fixes necessary to have a complete working production build via webpack. 

We cannot continue to dump assets into a separate folder, instead, any time a static asset needs to be linked via the app, it needs to be required within the component using it.

Example:
```
import { Component } from '@angular/core';

@Component({
  selector: 'navbar',
  templateUrl: './navbar.component.html'
})
export class NavbarComponent {
    public logoUrl: string = require('./climate-lab.svg');
}
```

### Demo

Serving the application via node's http-server on port 3101:
cclab-prod-build

### Notes

We will need to update the webpack prod config with the hostname of wherever we end up deploying this application. 

I split the webpack config so that the dev env builds to `build/` and the prod env builds to `dist/`, because there were issues with prod builds with the dev server running.

The production build is ~2mb. I'm not too concerned with this ATM since this is primarily designed as a desktop application. However, angular 2 supports things like build tree shaking to remove unused code paths, so we could investigate this as a later optimization.

Removed the 'Tech Stack' section in the README since it doesn't provide much value and was always out of date.

## Testing Instructions

- `npm run build:prod`
- `npm install -g http-server`
- `http-server -p 3101 dist/`
- Navigate to [http://localhost:3101/]()

Closes #53


fungjj92:
Nice +1

PR #65, opened 2016-09-14
Add a spinner

### LINKED ISSUE #55: Add chart loading spinner

flibbertigibbet:
Shows a spinner in place of the graph while data is loading.

Closes #55.

### DIFF
--- a/src/app/services/chart.service.ts
+++ b/src/app/services/chart.service.ts
@@ -79,8 +79,12 @@ export class ChartService {
 
     loadChartData(): void {
         let queries = [];
-
         let options = this.dataQueryOptions;
+
+        // send an empty notification to indicate that data is currently loading
+        if (this.chartDataObserver) {
+            this.chartDataObserver.next([]);
+        }
 
         // query like:
         // https://staging.api.futurefeelslike.com/api/climate-data/1/RCP85/?variables=pr&years=2050:2051
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -31,6 +31,7 @@ import { LoginComponent } from './login/login.component';
 import { NavbarComponent } from './navbar/navbar.component';
 import { PageNotFoundComponent } from './http-status/page-not-found/page-not-found.component';
 import { SidebarComponent } from './sidebar/sidebar.component';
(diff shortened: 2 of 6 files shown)


src/app/services/chart.service.ts
```diff
@@ -79,9 +79,13 @@ export class ChartService {
 
     loadChartData(): void {
         let queries = [];
-
         let options = this.dataQueryOptions;
 
+        // send an empty notification to indicate that data is currently loading
+        if (this.chartDataObserver) {
+            this.chartDataObserver.next({});
```
CloudNiner: Looks like next should be passed an `[]` instead of `{}` to satisfy the type checker

CloudNiner:
I think I figured out why you were unable to import `WaveComponent` from the spin kit node_module. 

It looks like that project declares `@angular/core` as a dependency, rather than a peer dependency. Meaning that when you import the file from node_modules, the old angular version becomes looped into the dependency graph, and that component is created using an older version of the `@Component` decorator. I tested this by manually creating a NgModule in the node_modules/ng2-spin-kit dir and attempted to import that instead. Which fails by stating that `@NgModule` doesn't exist in `@angular/core` which would be expected if the older angular version were being used. 

So, I guess copying the source files (assuming a valid license) and adding them to our project that way is the best we can do for now. It's probably worth a comment in the source file in our project stating that this is only temporary until the upstream project is updated to support RC6+. Or, we could submit a PR that fixes the upstream project. I think it would be pretty trivial.

CloudNiner:
LGTM

(I totally messed up my attempt at using the new review UI...oops)

flibbertigibbet:
Ok, yeah. Forgot to add attribution for the spinner library. It's MIT-licensed; mostly it is itself copy of the spinkit CSS. I did wind up modifying how it handles timeouts, as it would display anyway if it were initially set to be hidden.

src/app/services/chart.service.ts
flibbertigibbet: Looks like the new review tool doesn't pick up pushed changes. Done.

CloudNiner:
Heh. I saw that. I feel like the default `isRunning=true` isn't a great choice. But maybe theres some use case I'm not recognizing.

Issue #47, opened 2016-09-07
Adding a single chart re-queries for all data


flibbertigibbet:
Adding a chart to the list should only query the API for data for the added chart.

Currently it re-queries and reloads all charts.

CloudNiner:
Consider refactoring ChartComponent to handle its own data, rather that receiving it from ChartsComponent

fungjj92:
So that sounds totally sensible to me and I support it however I tried my hand at it and it's tricky.

Because there are some situations where we want to requery all of the charts, i.e. changing the dropdowns, and not, i.e. add/remove chart. If the data is queried at the chart level that also means there is not a unified place to signal global requerying. Though now that I think of it, it may be as easy as feeding the global vars as inputs through to chart level and relying on input change detection. 

In this vain, I think chart removal will be handled the same since chartList is still maintained in the same place (chart service). See #66.

flibbertigibbet:
The chart service should be able to handle the change fine. Re-querying for all should be a simple matter of adding a convenience function to the chart service to re-query for each. As there are no data dependencies across charts, there is no reason for us to delay processing until all chart queries return as we do now, except that we return a single, monolithic data object instead of one for each chart.

PR #66, opened 2016-09-14
Hook up remove chart button

### LINKED ISSUE #43: Add 'X' button next to settings to remove chart from list view

fungjj92:
## Overview

Hooks up the remove chart button ("x").

### Notes

I don't love this. I wanted chartList to be an observable that can be subscribed to (thus, no multiple calls to ```chartService.get```). I couldn't figure out how to do it though since we're not async calling data and observables in RxJS don't support watching vars or iterables (i.e. [knockout.js for example](http://knockoutjs.com/documentation/observableArrays.html)). It didn't seem necessary or desirable to use the aforelinked library, knockout.js.

Closes #43

### DIFF
--- a/src/app/charts/chart.component.html
+++ b/src/app/charts/chart.component.html
@@ -19,7 +19,8 @@
                 tooltipTrigger="mouseenter"
                 tooltipPopupDelay="300"
                 title=""
-                tooltip="Remove"><i class="icon icon-cancel"></i>
+                tooltip="Remove"
+                (click)="removeChart(indicator)"><i class="icon icon-cancel"></i>
         </button>
     </div>
 </div>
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
:+1: 

I'm thinking we should refactor (in another PR) so charts hold their own data, instead of the chart container holding the data for all charts. That would allow us to do data-dependent chart things like [load them separately](https://github.com/azavea/climate-change-lab/issues/47), and display loading spinners for them independently of one another. That should also let us remove a chart directly, without having to fire an event.

CloudNiner:
+1 to the refactor idea!

fungjj92:
I agree, I left a comment on #47 with continued thoughts

Issue #69, opened 2016-09-16
Graph control toggles don't persist on graph update


fungjj92:
Feature being on/off is maintained but the toggle default resets to off if models, scenario, or city change.

<img width="542" alt="screen shot 2016-09-16 at 11 13 37 am" src="https://cloud.githubusercontent.com/assets/10568752/18590938/a23a30a4-7bfe-11e6-9c7d-7c4d6548b828.png">

CloudNiner:
Closed by #70 and #72

PR #70, opened 2016-09-19
Feature/serialize ui state

### LINKED ISSUE #54: Serialize lab view state to json in localstorage

CloudNiner:
## Overview

Update the Lab application to serialize project state to window.localStorage when the LabComponent is destroyed

### Notes

This is a pretty big refactor, had to touch alot of the internals.

Items of note:
- Rmoved ChartContainerComponent as it was an unnecessary layer of indirection that wasn't used for anything
- Refactor querying for ClimateModel/Scenario to separate services and simplified the public API
- Added the Chart and Project view models, and made their respective components use these instead of properties on the component (this made serialization much easier)
- Moved the chart loading spinner so that the chart settings div remains visible (only the chart body where the drawing happens is replaced by the spinner)

Future TODOs (all of these are separate tasks I think):
- Figure out how to get project to save if page is refreshed (state is only saved currently when the LabComponent is destroyed, e.g. on a state change back to dashboard). I don't know what the hook for this is.
- Make each chart load its own data. This should be easier now too with the removal of ChartContainerComponent and the refactor to use a Chart model class.
- Refactor ChartService so that it uses a ClimateModel and Scenario property rather than saving strings (which doesn't take advantage of typings). Probably makes sense to do this in combination with the previous TODO

## Testing Instructions

 * Restart dev server
 * Add a few indicators, set some options on them
 * Return to dashboard
 * Click back into the Lab view from the dashboard -- All previous state should be saved

Closes #54 and #69


src/app/models/project.ts
```diff
@@ -0,0 +1,52 @@
...
+        };
+    }
+
+    static fromJSON(object: Object) {
+        return new this(object);
+    }
+
+    private generateUUID(): string {
+        // TODO: Replace with an actual implementation
+        // TODO: Remove when UUID generated via POST to API
```
flibbertigibbet: I'm confused by the two `TODO`s here: if the POST will return the ID, why would we need to generate a UUID client-side?

src/app/services/project.service.ts
```diff
@@ -0,0 +1,41 @@
+import { Injectable } from '@angular/core';
+
+import { Project } from '../models/project';
+
+@Injectable()
+export class ProjectService {
+
+    private storageKey: string = 'cclab.projects';
+    private projects: Project[] = [];
+    private storage: Storage = window.localStorage;
```
flibbertigibbet: Not sure how `window` is supposed to be handled in Angular 2, but seems like we probably shouldn't be referencing it directly here. Should it be [injected](http://stackoverflow.com/questions/34177221/angular2-how-to-inject-window-into-an-angular2-service) instead? How could we mock this for test?

src/app/services/scenario.service.ts
```diff
@@ -0,0 +1,32 @@
+import {Injectable} from '@angular/core';
```
flibbertigibbet: Inconsistent spacing around braces

fungjj92:
Breaks in firefox: need to request data as JSON

CloudNiner:
@fungjj92 added quick fix, pull down update and try now.

fungjj92:
To restore the scrolling, nest the charts in a div with class="scrollable" -- this was the class on chartComponent that made the div scroll

src/app/services/project.service.ts
CloudNiner: Yeah, it definitely should be injected. I attempted this, and it was more of a pain than I expected.

So I decided to defer, because this use of localStorage may not persist if we switch to having the API user account store the project information on the server.

src/app/models/project.ts
CloudNiner: Oops, forgot to do the first TODO, I'll throw something temporary in quick and just remove that first one.

CloudNiner:
Thanks, I'll fix scrolling as well

PR #72, opened 2016-09-21
Feature/issue 47 bugfix data requery

### LINKED ISSUE #36: Update ng2-bootstrap version

### LINKED ISSUE #47: Adding a single chart re-queries for all data

CloudNiner:
## Overview

Refactor the application so that each chart owns its own API requests. This meant mostly gutting the ChartService, which now just has the helper transform methods, and moving the Indicator API HTTP request to a more generic method on IndicatorService. See Notes for additional items of interest changed with this PR.

### Demo

cclab-chart-owns-data
cclab-new-modal

### Notes

- To simplify the climate model selection logic, replaced the checkbox with two buttons that select all/none. Removed the close button and switched to backdrop: static so that the only way to close the modal is to click the update button. Change vetted by @jbranigan 
- To handle re-adding the `backdrop: static` property of the modal, updated to ng2bootstrap 1.1.5 and ng2 2.0.0 release version
- Rather than storing the list of currently selected models in ChartService, `labComponent.project.models` is now the source of truth for the set of currently selected models. The `labComponent.climateModels` property is the list of all available models, and its selected property is kept in sync with the set of models in the `labComponent.project.models` array.
- ChartComponent now fires an event to indicate when one of its settings have changed, so that LabComponent can listen and save the state to localStorage. This fixes the issue where some parts of the UI state would not persist if the page was refreshed.
- The last commit removes a bunch of unused imports throughout the project

## Testing Instructions

 * Stop dev server
 * npm install && npm start
 * Add a few charts from the sidebar, each should independently load data
 * Deleting a chart no longer requeries data for other charts
 * Change anything in the UI, do a full page refresh, state should remain consistent

Note: The `defaultYears` property in `constants.ts` can be safely deleted, it is no longer used.

Closes #36, Closes #47


src/app/lab/lab.component.ts
```diff
@@ -26,26 +26,43 @@ import * as _ from 'lodash';
...
+  public scenarios: Scenario[] = [];
+  public project: Project;
+
+  constructor(private chartService: ChartService,
               private climateModelService: ClimateModelService,
               private scenarioService: ScenarioService,
               private projectService: ProjectService) {
-    // necessary to catch application root view container ref. see:
-    // https://valor-software.com/ng2-bootstrap/#/modals
-    this.viewContainerRef = viewContainerRef;
```
CloudNiner: Moved because it was in the wrong place, it needs to be in the constructor for the top-level component, which is AppComponent, not LabComponent.

src/app/lab/lab.component.ts
```diff
@@ -68,94 +85,71 @@ export class LabComponent implements OnInit, OnDestroy {
...
+  private getClimateModels() {
+    this.climateModelService.list().subscribe(data => {
+      this.climateModels = data;
+
+      // Initialize 'selected' attributes with models in project
+      if (this.project.models.length === 0) {
+        this.selectAllClimateModels();
+        this.updateProjectClimateModels();
+      } else {
+        this.project.models.forEach(projectModel => {
```
CloudNiner: Just brute forced this update/comparison. Always a very small dataset so I don't think its worth it to optimize.

CloudNiner:
Ready for review.

src/app/charts/line-graph.component.ts
```diff
@@ -238,7 +238,7 @@ export class LineGraphComponent {
 
                 // Prepare data for bar graph
                 let maxBars = _.map(this.extractedData, (datum, index) => {
-                    let val = this.minVal <  this.yData[index] ? y * 2 : 0;
+                    let val = this.maxVal <  this.yData[index] ? y * 2 : 0;
```
fungjj92: My b, I merged a fix for this

src/app/charts/chart.component.ts
```diff
@@ -1,34 +1,56 @@
...
+                private indicatorService: IndicatorService) {}
+
+    ngOnChanges() {
+        this.chartData = [];
+        this.indicatorService.getData({
+            indicator: this.chart.indicator,
+            scenario: this.scenario,
+            city: this.city,
+            climateModels: this.models
+        }).subscribe(data => this.chartData = this.chartService.convertChartData([data]));
```
fungjj92: I believe we should ngOnDestroy this subscription

src/app/lab/lab.component.ts
```diff
@@ -26,26 +26,43 @@ import * as _ from 'lodash';
...
 
-  private chartData: ChartData[];
-  public scenarios: Scenario[];
+    this.getClimateModels();
+    this.getScenarios();
+  }
 
-  public viewContainerRef: ViewContainerRef;
+  ngOnDestroy() {
+    this.projectService.update(this.project);
```
fungjj92: For convention, I think we should also unsubscribe from observables here. Memory leak is not a risk here because I think this component lives for the length of the whole app, as exemplified: https://angular.io/docs/ts/latest/cookbook/component-communication.html#!#bidirectional-service

CloudNiner: This component does not live for the length of the app. It gets destroyed and recreated every time you load the `/lab` route. I'll look into unsubscribing the observables.

CloudNiner:
Ok. After looking at this: http://stackoverflow.com/questions/35042929/do-you-need-to-unsubscribe-from-angular-2-http-calls-to-prevent-memory-leak

ng2 HTTP observables do not need to be unsubscribed from. I did remove the cached observable in the ClimateModelService and the ScenarioService, commit log for that copied here for clarity:
```
The HTTP service automatically cleans up it's http request
observables by calling observable.complete() when the request
returns OK. By saving creating a new observable and returning
that, we remove this protection, and potentially add bugs
where multiple components subscribed to the same observable
would all update together if a new request is made.
```

fungjj92:
LMK if you see the blacking out or not. Also, I am still figuring out how to use this new github review mode. The PR looks good to me aside from this new blacking out.

CloudNiner:
Yeah, I do see the blacking out. Fixed by adding missing "fade" animation to a new _animations.scss file, copied from the bootstrap source.

Also added some checks to prevent the UI from getting to a state where no models are selected.

Issue #86, opened 2016-09-27
City autocomplete stackoverflows if no city 


fungjj92:
<img width="627" alt="screen shot 2016-09-27 at 2 00 06 pm" src="https://cloud.githubusercontent.com/assets/10568752/18885697/bc494542-84ba-11e6-9968-82d02ecd61c4.png">

If you have a project with no city and also have the city autocomplete present, the autocomplete stack overflows with the above error
Discovered by the add/edit project workflow

Issue #85, opened 2016-09-27
Model modal submits on add/edit project view


fungjj92:
It shouldn't submit the form when you click 'update' models, just close the modal and add selected models to the active project object.

Issue #79, opened 2016-09-22
Edit project modal


CloudNiner:
Add 'project settings' link in navbar dropdown to re-open project setup modal to allow editing of name/description

CloudNiner:
Depends on #82, reuse the same modal component and add an editing mode

PR #87, opened 2016-09-27
Feature/new & edit project workflow

### LINKED ISSUE #80: Implement new project workflow

fungjj92:
## Overview

Create or edit a project and save all, too! It can accept projects with no input.

The edit project button only shows up when you're in a project.

### Demo

screen shot 2016-09-27 at 2 19 47 pm
screen shot 2016-09-27 at 2 20 08 pm
screen shot 2016-09-27 at 2 21 10 pm
screen shot 2016-09-28 at 10 13 50 am

### Notes

Rather than modals I made add/edit project into fullpage views. Flexible.
Light styling.
When selecting models, if you click "update" it will submit the add/edit form which is not desired. Opened issue #85.

## Testing Instructions

You can only access the lab with a project id in the URL, now.

Create a project from the dashboard and it should lead you directly to it.
Else, localhost:3100/lab/your_existing_project_id_here

Edit: from lab/project use the navbar user dropdown, project settings. OR localhost:3100/edit/project_id_here

You can find your projects by using Augury debugging and looking into your projectService which stores an array of projects.

Closes #80, #79, #86
Opens issues #85


CloudNiner:
Happy to take a look at this if its ready for review.

fungjj92:
Now ready. I added a view for edit project.

CloudNiner:
Is there a reason not to look at #86 as part of this PR? I ran into trouble with that error until it randomly worked. And I'm not sure why.

src/app/lab/dropdowns/city-dropdown.component.ts
```diff
@@ -0,0 +1,67 @@
...
+                    placeholder="Enter your city"
+                    min-chars="2" />
+              </div>
+            </div>`
+})
+export class CityDropdownComponent {
+
+    public apiCities: string = apiHost + "/api/city/?search=:keyword";
+
+    @Input() project: Project;
```
CloudNiner: Shouldn't this component take a City object, rather than Project?

src/app/lab/dropdowns/model-modal.component.html
```diff
@@ -0,0 +1,56 @@
...
+    aria-haspopup="true" aria-expanded="true"
+    (click)="smModal.show()">
+    <i class="icon-flash"></i>
+    {{ selectedClimateModels }}
+    <label>
+      <a>Select Models</a>
+    </label>
+  </button>
+</div>
+<!-- TODO: remove config option to not display modal background once issue fixed:
```
CloudNiner: Forgot to remove this TODO, it ca be removed here.

src/app/lab/dropdowns/model-modal.component.ts
```diff
@@ -0,0 +1,80 @@
...
+
+@Component({
+  selector: 'model-modal',
+  templateUrl: './model-modal.component.html'
+})
+export class ModelModalComponent implements OnInit {
+
+    @Input() project: Project;
+
+    public climateModels: ClimateModel[] = [];
```
CloudNiner: I think this is another component that could use better scoping of its inputs/outputs. Rather than passing the entire project, this component only needs the list of selected models to set its initial state. Then, if its output was something like: `onModelsSelected(models: ClimateModel[])` which fired when the modal was closed, you could also easily clean up #85 by handling this event differently in the LabComponent and AddEditProjectComponent.

src/app/lab/dropdowns/scenario-dropdown.component.ts
```diff
@@ -0,0 +1,56 @@
...
+                    tooltipPlacement="bottom"
+                    tooltipTrigger="mouseenter"
+                    tooltipPopupDelay="300">{{ scenario.name }}</a>
+                </li>
+              </ul>
+            </div>`
+})
+export class ScenarioDropdownComponent implements OnInit {
+
+    @Input() project: Project;
```
CloudNiner: Same thing, why not just input the scenario?

src/app/lab/lab.component.ts
```diff
@@ -2,21 +2,19 @@
...
  */
-import { Component, OnInit, OnDestroy, ViewEncapsulation } from '@angular/core';
+import { Component, OnInit, OnDestroy, ViewEncapsulation, Input } from '@angular/core';
+import { Router, ActivatedRoute }     from '@angular/router';
+
 import { SidebarComponent } from '../sidebar/sidebar.component';
-import { ChartService } from '../services/chart.service';
-import { ClimateModelService } from '../services/climate-model.service';
-import { ScenarioService } from '../services/scenario.service';
+import { ScenarioDropdownComponent } from './dropdowns/scenario-dropdown.component';
```
CloudNiner: No need to import these here, they're available already because they're defined in the app module.

src/app/lab/lab.component.ts
```diff
@@ -27,95 +25,33 @@ import * as _ from 'lodash';
...
-      this.projectService.add(this.project);
-    }
-
-    this.getClimateModels();
-    this.getScenarios();
+    // Load existing project id or redirect to dashboard
+    this.route.params.subscribe(params => {
+      let id: string = params['id'];
+      if (id !== undefined) {
+        this.project = this.projectService.get(id);
```
CloudNiner: Probably also want to guard against the case where projectService.get is passed an invalid ID and doesn't find a matching project, and navigate back to dashboard in that case as well.

src/app/models/project.ts
```diff
@@ -13,8 +13,8 @@ export class Project {
     name: string;
     description: string;
     visibility: ProjectVisibility = ProjectVisibility.Private;
-    city: City;
-    scenario: Scenario;
+    city: City = {} as City;
```
CloudNiner: What's the reason for this change? While this might be work, an empty object for City/Scenario is not valid (these types have required parameters) and you're forcing the type system to accept a value that isn't actually a City/Scenario, which could lead to issues down the road. There's probably a better way depending on what the issue you're trying to get around is.

src/app/navbar/navbar.component.html
```diff
@@ -22,6 +22,10 @@ <h1 class="h4 logo">
             <a class="dropdown-item" [routerLink]="['/']">Go to Dashboard</a>
           </li>
           <li class="divider dropdown-divider"></li>
+          <li role="menuitem" *ngIf="projectID">
+            <a class="dropdown-item" [routerLink]="['/edit/', projectID]">Project Settings</a>
+          </li>
+          <li class="divider dropdown-divider"></li>
```
CloudNiner: Probably also want to `*ngIf="projectId"` this divider. Otherwise on the dashboard page you have two dividers in the dropdown, separating nothing.

src/app/project/add-edit-project.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    constructor(private router: Router,
+                private route: ActivatedRoute,
+                private projectService: ProjectService) {}
+
+    ngOnInit() {
+        // Load current project
+        this.route.params.subscribe(params => {
+            let id: string = params['id'];
+            if (id !== undefined) {
+                this.model.project = this.projectService.get(id);
```
CloudNiner: Here's another place to handle an invalid project id.

src/app/project/add-edit-project.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    public edit: boolean = false;
+    public model = {'project': {} as Project};
+
+    constructor(private router: Router,
+                private route: ActivatedRoute,
+                private projectService: ProjectService) {}
+
+    ngOnInit() {
+        // Load current project
+        this.route.params.subscribe(params => {
```
CloudNiner: Is this an observable subscription that needs to be unsubscribed?

fungjj92:
Re: the multiple cases of passing the whole project rather than specific properties, we are given listening for changes to the project and the app updates accordingly. I had thought about this too, but upon trial of passing the specific properties found we'd then have to set up event emitters, which would be messier.

src/app/models/project.ts
fungjj92: Yea I couldn't figure out the elegant place for this.
Essentially loading a project with empty city, scenario fields was problematic in the add project workflow. This is a cause of expectations of autocompletes and dropdowns. I can set these fields to {} there, rather than in the model.

CloudNiner:
Re: passing around project. I do think it would be a bit more messy. In a more complicated project we'd probably want to be more careful about what we require as I/O for components, but you're probably right in that its ok for now. We can revisit later if it becomes an issue.

CloudNiner:
Should I be able to click the 'create project' button if i haven't yet input anything into any of the form fields? It seems to me that at least a city and scenario should be required on the new project page, as that's what's required for the API calls to Indicator to work.

CloudNiner:
Probably should still handle invalid project ids being passed to the ProjectService in the places noted, but other changes LGTM.

fungjj92:
Yeah, I still have to handle the model modal and invalid ID. They stumped me yesterday. I'll let you know when those are ready!

src/app/lab/lab.component.ts
```diff
@@ -23,43 +23,47 @@ import * as _ from 'lodash';
...
+    ngOnInit() {
+        // Load existing project id or redirect to dashboard
+        this.subscription = this.route.params.subscribe(params => {
+            let id: string = params['id'];
+            if (id !== undefined) {
+              this.project = this.projectService.get(id);
+            }
+        });
+        // Reroute to dashboard if project doesn't exist
+        if (!this.project) {
```
CloudNiner: I don't think this is doing what you expect. Shouldn't this check be inside the subscribe handler right after the project is pulled from the ProjectService?

Also consider moving all of the logic in this subscribe onNext handler to a separate named function as a stylistic improvement.

src/app/project/add-edit-project.component.ts
```diff
@@ -39,14 +41,20 @@ export class AddEditProjectComponent implements OnInit {
...
-            this.model.project.scenario = {} as Scenario;
-            this.model.project.city = {} as City;
+            this.model = new ProjectForm(new Project({
+                // Dropdowns & autocompletes require a value, {} at the very least
+                'scenario': {} as Scenario,
+                'city': {} as City
+            }));
+        }
+        // Reroute to dashboard if project doesn't exist
+        if (!this.model.project) {
```
CloudNiner: Same thing here. Shouldn't this check be within the subscribe handler after the project is pulled?

CloudNiner:
Almost there! Double check the logic for the params rerouting, I put something like http://localhost:3100/#/lab/projectdoesntexist into my url bar and i just got a blank screen, rather than being redirected.

fungjj92:
That is weird! I definitely get redirected to the dashboard....Show me tomorrow??

PR #92, opened 2016-10-04
Closes #71: Fix threshold bands for below zero values

### LINKED ISSUE #71: Threshold bars are weird if y axis includes negative values
When the feature was written we only had mock data whose values were all > 0. 

If the y scale includes 0 and below, the threshold feature freaks out (top: bad, bottom: good):
screen shot 2016-09-20 at 12 51 26 pm

General clean up of bar heights necessary all around, too.

sharph:
## Overview

This fixes the broken visualization of data above or below thresholds as described in issue #71. It also specifies a fixed height to match the charting area.

### Demo

bands

## Testing Instructions

* Choose a city with indicators with negative numbers. Anchorage, AK, for example.
* Pick an indicator with negative numbers (Yearly Minimum Low Temperature)
* Make sure threshold display does not look broken

Closes #71

### DIFF
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -216,34 +216,34 @@ export class LineGraphComponent {
             let x2 = this.xRange[0];
             let y = D3.max(this.yData);
 
-            if (this.min && this.minVal) {
+            if (this.min) {
                 // Draw min threshold line
                 let minData = [{'date': x1, 'value': this.minVal}, {'date': x2, 'value': this.minVal}];
                 this.drawLine(minData, 'min-threshold');
 
                 // Prepare data for bar graph
-                let minBars = _.map(this.extractedData, (datum, index) => {
-                    let val = this.minVal >  this.yData[index] ? y * 2 : 0;
-                    return { 'date': datum['date'], 'value': val };
-                });
+                let minBars = _(this.extractedData)
+                              .map((datum, index) => ({ 'date': datum['date'] }))
+                              .filter((bar, index) => this.minVal > this.yData[index])
+                              .value();
 
                 // Add bar graph
-                this.drawBarGraph(minBars, 'min-bar');
+                this.drawBands(minBars, 'min-bar');
             }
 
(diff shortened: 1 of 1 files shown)


src/app/charts/line-graph.component.ts
```diff
@@ -270,4 +270,20 @@ export class LineGraphComponent {
           .attr('y', d => this.yScale(d.value))
           .attr('height', d => this.height - this.yScale(d.value));
```
fungjj92: drawBarGraph isn't used anymore so can be deleted

src/app/charts/line-graph.component.ts
```diff
@@ -216,34 +216,34 @@ export class LineGraphComponent {
...
+                              .filter((bar, index) => this.minVal > this.yData[index])
+                              .value();
 
                 // Add bar graph
-                this.drawBarGraph(minBars, 'min-bar');
+                this.drawBands(minBars, 'min-bar');
             }
 
-            if (this.max && this.maxVal) {
+            if (this.max) {
```
fungjj92: For this, we'd need to keep `&& this.maxVal` and `&& this.minVal` or the dotted line shows up and in sometimes undesirable places (e.g. below graph)

sharph: Having `&& this.maxVal` had the effect of hiding the bars when `maxVal` was set to zero, as it is falsey. If I leave `&& this.maxVal` in, the line still shows in inappropriate places when it is negative.

This should really be comparing `maxVal` and `minVal` against the `y` range of the chart. I'll look into adding this.

sharph: Addressed in https://github.com/azavea/climate-change-lab/pull/92/commits/1f026f58154df5c52a9000144b07bea60ba83c1f

sharph:
Ready for another look!

src/app/charts/line-graph.component.ts
fungjj92: Nice

fungjj92:
+1

PR #94, opened 2016-10-05
Add my projects view and bootstrap

### LINKED ISSUE #82: Implement 'My Projects' view
Store data in localstorage

https://marvelapp.com/h584hg/screen/14807008

fungjj92:
## Overview

Creates a paginated projects view page.
Also added classic bootstrap stylesheet and cleaned up accordingly

### Demo

screen shot 2016-10-05 at 3 15 30 pm

### Notes

Styling isn't quite right on the table, did what I could.

## Testing Instructions

A page holds 5 projects (arbitrary). Add projects. Delete them. 

Closes #82

### DIFF
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -1,18 +1,28 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
 
 import { Project } from '../models/project';
+import { ProjectService } from '../services/project.service';
 
 @Component({
   selector: 'dashboard',
   templateUrl: './dashboard.component.html'
 })
-export class DashboardComponent {
+export class DashboardComponent implements OnInit {
 
-    // TODO: replace any type with a project model
     private projects: Project[];
 
-    constructor() {
-        // TODO: Eventually query for saved projects
-        this.projects = [];
+    constructor(private projectService: ProjectService) {}
+
+    ngOnInit() {
+        this.getProjects();
+    }
+
+    public getProjects() {
+        this.projects = this.projectService.list();
+    }
+
+    public deleteProject(project) {
+        this.projectService.remove(project);
+        this.getProjects();
     }
 }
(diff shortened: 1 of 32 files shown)


src/app/dashboard/dashboard.component.ts
```diff
@@ -1,18 +1,49 @@
...
   templateUrl: './dashboard.component.html'
 })
-export class DashboardComponent {
+export class DashboardComponent implements OnInit {
 
-    // TODO: replace any type with a project model
     private projects: Project[];
+    private pageOfProjects: Project[];
+    public currentPage: number = 1;
+    public itemsPerPage: number = 5;
```
sharph: Might be good to use `const` here

src/index.html
```diff
@@ -5,6 +5,7 @@
   <meta http-equiv="x-ua-compatible" content="ie=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,800" rel="stylesheet">
+  <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" rel="stylesheet">
```
sharph: In the past, we've imported this as SCSS into the project. Was that considered?

CloudNiner: Hmmm, correct. I don't think this is the best way to accomplish this. I thought the ng2-bootstrap components contained their own styles? 

If this is included to get the list displaying properly, that's something that could better be handled by flexbox and the existing classes in the project I think. @jfrankl do you have thoughts on this?

fungjj92: I thought ng2-bootstrap contained styles too, but as it turns out it is just the bootstrap features as components, sans CSS. The homepage says that for CSS, and they cited the link http://valor-software.com/ng2-bootstrap/index-bs4.html#/

I think we will want bootstrap generally, but certainly lmk what direction we want to go in!

src/app/dashboard/dashboard.component.ts
fungjj92: Hm a would-be good suggestion, but with const the compiler complains, "A class member cannot have the 'const' keyword," so I'm leaving it as is.

src/index.html
CloudNiner: Ok cool. I just took the style cleanup task for this card, so I'll work from this branch and put up a PR once I have something.

CloudNiner:
Think this is ready to go...looks great!

PR #100, opened 2016-10-07
Feature/my projects style

### LINKED ISSUE #95: My Project view cleanup
Cleanup styling
- Reduce header whitespace
- Fix table spacing and justification

Add 'last updated' column

See https://marvelapp.com/h584hg/screen/14807008

CloudNiner:
## Overview

Cleans up style of add edit project view and my projects list on dashboard.

Uses the mantle framework as much as possible, but also includes bootstrap via node_modules in case we need it in the future.

### Demo

cclab-add-edit-styles
cclab-myproject-styling

### Notes

- Replaced the bootstrap paginator for custom links that match the style of the marvel app wireframes.

Closes #95

### DIFF
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -18,7 +18,7 @@ export class DashboardComponent implements OnInit {
 
     ngOnInit() {
         this.getProjects();
-        this.setPage();
+        this.setPage(this.currentPage);
     }
 
     public getProjects() {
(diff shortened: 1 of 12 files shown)


src/app/dashboard/dashboard.component.ts
```diff
@@ -18,31 +18,25 @@ export class DashboardComponent implements OnInit {
...
-            this.projectService.remove(project);
-            this.getProjects();
-        } else {
-            this.projectService.remove(project);
-            this.getProjects();
-            this.setPage();
+        this.projectService.remove(project);
+        this.getProjects();
+        let page = this.currentPage;
+        if (this.projects.length % this.itemsPerPage === 0) {
```
fungjj92: If you have n* this.itemsPerPage +1 projects (i.e, 6, 11, 16..) and delete a project on a full page, this check fails us and sends us to the previous page (which in the case of 6 -> projects, sends up to blankness)

Oddly enough any deleting any project always produces this error on Firefox.

fungjj92:
Otherwise nice +1

src/app/dashboard/dashboard.component.ts
CloudNiner: Whoops! You're totally right...great catch. Will fix.

PR #103, opened 2016-10-17
Single-chart data scrubber

### LINKED ISSUE #48: Implement scrubber to view data value at a given X
cclab-data-scrubber

fungjj92:
## Overview

Individual chart data scrubbers!

### Demo

**Prototype:**
https://cloud.githubusercontent.com/assets/1818302/18316520/f0f13a3e-74e8-11e6-86ab-8cd037dbdd68.png

**Demo:**
screen shot 2016-10-17 at 5 25 41 pm

### Notes

Events are tricky, especially between browsers. I tried all sorts of ways in all sorts of places.
1) **D3's built in svg event handler** ( e.g, your_svg.on('mouseover')) -- crucial event information not recognized in Firefox.
2) **JQuery events** -- Event information avail across browsers however this method resulted in event listeners expiring after use, which is not useful. Also, JQuery.
3) **Angular HostListeners** --most functional solution I have now. 

You will see there is a host listener at chart-component level and another at line-graph-component level. I am not certain how expensive it is to have listeners in both places but doing so was necessary in practice. Initial, failed, strategy was there would just be a listener at line-graph-component level only. Of benefit, I think this structure helps set up the cross-chart scrubber.

**D3 & Tooltip**
To the best of my figuring, d3-tooltip would not work here here combined with the moving mouseover scrubber. This unusual alternative is a bunch of layered SVG elements created and destroyed based on mouse events.

We will want to adjust unit options here or on the API side. For example, "count" is not a unit and labelling "15 count" is awkward.

Closes #48

### DIFF
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -140,7 +140,7 @@ export class LineGraphComponent {
         const maxY = D3.max(_.map(this.extractedData, d => d.values.max));
         const yPad = (maxY - minY) > 0 ? (maxY - minY) * 1/3 : 5; // Note: 5 as default is arbitrary
         // if minY is 0, keep it that way
-        this.yScale.domain([minY == 0? minY: minY - yPad, maxY + yPad]);
+        this.yScale.domain([minY == 0 ? minY : minY - yPad, maxY + yPad]);
 
         // Expects line data as DataPoint[]
         this.valueline = D3.line()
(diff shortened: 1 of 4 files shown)


src/app/charts/line-graph.component.ts
```diff
@@ -279,26 +296,106 @@ export class LineGraphComponent {
...
+            .attr('width', this.width);
+
+        // Toggle scrubber visibility
+        this.hover? $('.' + this.id).toggleClass('hidden', false) : $('.' + this.id).toggleClass('hidden', true);
+    }
+
+    private redrawScrubber(event) {
+        let xPos = event.offsetX - this.margin.left;
+        // Firefox handles event positioning differently than Chrome, Safari
+        if (navigator.userAgent.indexOf('Firefox') != -1) {
```
CloudNiner: Might want to change the check to `if (navigator && navigator.userAgent.toLowerCase().indexOf('Firefox') !== -1)`. These changes allow:
- The script to not throw an exception if the navigator global is missing (i.e. in tests)
- A case insensitive check on the UA string (which is a bit less specific)
- Triple-equals which is more correct and gets you into the habit of using triple equals by default for exact type equality

PR #104, opened 2016-10-20
Feature/multi chart scrubber

### LINKED ISSUE #97: Prototype a cross-chart scrubber
Mousing over one chart shows value at same year across all charts

Only apply to yearly aggregation charts for now

https://marvelapp.com/h584hg/screen/14807022

fungjj92:
## Overview

Multi chart scrubber for yearly indicators!

### Demo

screen shot 2016-10-20 at 11 30 24 am

screen shot 2016-10-20 at 11 55 26 am

screen shot 2016-10-20 at 11 36 12 am

### Notes

Multi-scrubber depends upon 2 Observable streams. One listens for any chart #overlay being moused over and relays it to every line graph component. The other listens for mouse positioning and relays that to every line graph component to redraw the scrubber.

## Testing Instructions

Toggle multi-chart scrubber in project settings and mouse over charts.

Closes #97

### DIFF
--- a/src/app/services/chart.service.ts
+++ b/src/app/services/chart.service.ts
@@ -1,5 +1,5 @@
-import {Injectable} from '@angular/core';
-import 'rxjs/Rx';
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs/Subject';
 
 import { ChartData, MultiDataPoint } from '../models/chart';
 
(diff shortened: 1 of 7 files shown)


CloudNiner:
Won't have a chance to look at this before RRP, feel free to reassign as necessary.

fungjj92:
That's okay, no rush on it.

src/app/services/chart.service.ts
```diff
@@ -13,8 +13,23 @@ import * as _ from 'lodash';
 @Injectable()
 export class ChartService {
 
+    private multiChartScrubberInfo = new Subject();
+    private multiChartScrubberHover = new Subject<Boolean>();
+
+    multiChartScrubberInfo$ = this.multiChartScrubberInfo.asObservable();
```
CloudNiner: Is this `$` postfix a common convention in the rxjs world? If it isn't, it may be better to just make the public property just be `multiChartScrubberInfo` or `multiChartScrubberObservable` and then give the private var the same name with an `_` prefix in order to be consistent with existing JS naming conventions.

src/app/services/chart.service.ts
```diff
@@ -13,8 +13,23 @@ import * as _ from 'lodash';
 @Injectable()
 export class ChartService {
 
+    private multiChartScrubberInfo = new Subject();
```
CloudNiner: A subject seems pretty correct for this implementation. While I have limited experience as well, Subjects can be frowned upon as they have state, which is no good for the functional programming paradigm, and so see limited use.

I found this, appears to be a great deep dive into when to and when to not use Subjects:
http://davesexton.com/blog/post/To-Use-Subject-Or-Not-To-Use-Subject.aspx

src/app/services/chart.service.ts
rmartz: Looks like it's a pattern [from Cycle.js](https://cycle.js.org/basic-examples.html#what-does-the-suffixed-dollar-sign-mean), I've seen it used in a couple places but it seems to stem from them.

PR #107, opened 2016-10-21
Support monthly indicators in line graphs

### LINKED ISSUE #105: Add support for monthly indicators

fungjj92:
## Overview

Line graphs weren't set up to take monthly indicator data. Now they are!

### Demo

screen shot 2016-10-21 at 11 51 05 am

Closes #105

### DIFF
--- a/src/app/charts/line-graph.component.ts
+++ b/src/app/charts/line-graph.component.ts
@@ -40,14 +40,12 @@ export class LineGraphComponent implements OnInit, OnDestroy {
     private yScale;                        // D3 scale in Y
     private htmlElement;                   // Host HTMLElement
     private valueline;                     // Base for a chart line
-    private xRange: Array<string>;         // Min, max date range
+    private xRange: Array<Date>;           // Min, max date range
     private xData: Array<number>;          // Stores x axis data as integers rather than dates, necessary for trendline math
     private yData: Array<number>;          // Stores primary y axis data, multi-use
-    private trendData: Array<DataPoint>;   // Formatted data for the trendline
+    private timeFormat: string;            // Date formatting for x axis labels (e.g, '%Y-%m')
     private scrubber;                      // Lump of scrubber elements
     private scrubberLine;                  // Scrubber element, independent
-    private timeOptions: any;
-    private timeFormat: string;
     private id: string;                    // Randomly generated int # used to distinguish the chart for drawing and isolated chart scrubber
                                            // Not a perfect solution should the random int and indicator be the same
                                            // However, this is quite unlikely (1/10000, and even less likely by way of app use)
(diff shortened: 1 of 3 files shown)


src/app/charts/line-graph.component.ts
```diff
@@ -132,6 +131,12 @@ export class LineGraphComponent {
         // Time scales only recognize annual and daily data
         var parseTime = D3.timeParse(this.timeFormat);
         this.extractedData.forEach(d => d.date = parseTime(d.date));
+
+        // Sort data by date ascending
+        this.extractedData.sort(function(a, b) {return a.date - b.date;});
```
KlaasH: This creates a type checking error:
```
[default] /home/local/AZVA-INT/khoekema/projects/climate/climate-change-lab/src/app/charts/line-graph.component.ts:136:56 
    The left-hand side of an arithmetic operation must be of type 'any', 'number' or an enum type.
```
To me it seems like the problem goes back a little ways.  I.e.

- The `DataPoint` type defines its `date` attribute as a string [here] (https://github.com/azavea/climate-change-lab/blob/feature/SupportMonthlyIndicators/src/app/models/chart.ts#L40)
- But the `LineGraphComponent` overwrites it with a javascript Date object [here] (https://github.com/azavea/climate-change-lab/blob/feature/SupportMonthlyIndicators/src/app/charts/line-graph.component.ts#L133)

That seems like a bad situation.  If we want dates, we should convert much earlier (e.g. [here] (https://github.com/azavea/climate-change-lab/blob/feature/SupportMonthlyIndicators/src/app/services/chart.service.ts#L41) where the data is initially loaded).  That would require moving the config that's used to parse and interpret the date to somewhere where it can be shared, but that might be a good thing anyway.  Anything that's using `DataPoint` values as defined (i.e. treating the dates as strings) would have to be checked/revised to handle Date objects.

If we mostly want to keep the date keys as returned by the API (i.e. 'YYYY', 'YYYY-MM', or 'YYYY-MM-DD' strings) then we should define a separate data type for use inside `LineGraphComponent` and explicitly convert the data to that format.  (Also, `setLineScales` seems like a weird place to mutate the data in a way that affects all subsequent steps. Doing it in `filterData` where the `extractedData` object is created would make more sense to me.)

fungjj92: I see your point. There isn't a reason to maintain the string format of `date` that comes from the API. It makes sense to process date in the service while we are chopping up the data.

I could define two types: RawDataPoint and DataPoint. Former, string date; Latter, date date.

fungjj92:
Updated. I am now processing the strings into dates in chart.service and we still only have 1 type, DataPoint which now expects a Date date not a string date. Should be no more linter errors.

src/app/charts/line-graph.component.ts
```diff
@@ -163,9 +153,11 @@ export class LineGraphComponent implements OnInit, OnDestroy {
 
     // Set axis and line scales
     private setLineScales(): void {
-        // Time scales only recognize annual and daily data
-        var parseTime = D3.timeParse(this.timeFormat);
-        this.extractedData.forEach(d => d.date = parseTime(d.date));
+        // Sort data by date ascending
+        this.extractedData.sort(function(a, b) {return +a.date - +b.date;});
```
KlaasH: Me: That seems like a kludge, can we do better?
The Google: https://github.com/Microsoft/TypeScript/issues/2361#issuecomment-82509802
Me: :scream:

Coercion it is!

PR #114, opened 2016-10-27
Closes #110: Add label property to Scenario and use in dropdown

### LINKED ISSUE #110: Format scenarios
The three scenarios should be formatted as follows:

image
- RCP 4.5
- RCP 8.5
- Historical

CloudNiner:
## Demo

screen shot 2016-10-27 at 16 06 52

### DIFF
--- a/src/app/lab/dropdowns/scenario-dropdown.component.ts
+++ b/src/app/lab/dropdowns/scenario-dropdown.component.ts
@@ -20,7 +20,7 @@ import { ProjectService } from '../../services/project.service';
                 id="scenarioDropdown" data-toggle="dropdown" aria-haspopup="true"
                 aria-expanded="true">
                 <i class="icon-flash"></i>
-                {{ project.scenario.name || "Select Scenario" }}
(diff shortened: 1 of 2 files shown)


src/app/services/scenario.service.ts
```diff
@@ -13,10 +13,22 @@ import { apiHost } from "../constants";
...
     constructor(private apiHttp: ApiHttp) {}
 
     public list(): Observable<Scenario[]> {
         let url = apiHost + '/api/scenario/';
-        return this.apiHttp.get(url).map(resp => resp.json() || [] as Scenario[]);
+        return this.apiHttp.get(url)
+            .map(resp => resp.json() || [] as Scenario[])
+            // Can't use Observable<any>.foreach() because that returns a promise
+            .map(s => {
+                s.label = this.LABELS[s.name] || s.name;
```
rmartz: Isn't label already a property of the `/api/scenario/` response?

CloudNiner: Derp. So it is. Updated.

fungjj92:
yay +1

PR #116, opened 2016-10-28
Feature/save projects to api

### LINKED ISSUE #96: Save Projects in ProjectService to API instead of localstorage

fungjj92:
## Overview

Now that Projects have been added to the API, lab is updated to actually save projects to the user account.
### Notes

There are some communication simplifications among chart changes and saving, which is a nice outcome.

3 ways that save: navigating away, update in 'project settings,' and X-ing out of a project.
## Testing Instructions

Try to break it or find an instance where your changes are not saved!

Closes #96


src/app/lab/lab.component.ts
```diff
@@ -24,22 +24,32 @@ import * as _ from 'lodash';
...
         this.routeParamsSubscription = this.route.params.subscribe(params => {
             let id: string = params['id'];
             if (id !== undefined) {
-              this.project = this.projectService.get(id);
-            }
-            // Reroute to dashboard if project doesn't exist
-            if (!this.project) {
+              this.projectService.get(id).subscribe(data => {
+                    this.project = data;
+                    if (!this.project) {
```
CloudNiner: If a project id doesn't exist, the API will throw a 404 which is handled by the observer `catch` method rather than the success method. I wouldn't expect data to ever be null/undefined here. Do the router navigate in the catch handler instead.

src/app/models/project.ts
```diff
@@ -8,8 +8,7 @@ export enum ProjectVisibility {
     Public
 }
 
-export class Project {
-    id: string;
+export class ProjectData {
```
CloudNiner: Only have one model class per file. Split this file into two named `project-data.ts` and `api-project.ts`

This makes it easy to always fuzzy search to find the model files, rather than having to guess which file the model is defined in.

src/app/models/project.ts
```diff
@@ -45,14 +40,18 @@ export class Project {
...
+}
+
+export class APIProject {
+    id: string;
+    created: string;
+    modified: string;
+    project_data: ProjectData;
 
-    private generateUUID(): string {
-        // TODO: Remove when UUID generated via POST to API
```
CloudNiner: Whoop.

src/app/services/project.service.ts
```diff
@@ -1,45 +1,43 @@
...
 export class ProjectService {
 
-    private storageKey: string = 'cclab.projects';
-    private projects: Project[] = [];
-    private storage: Storage = window.localStorage;
-
-    constructor() {
-        this.projects = JSON.parse(this.storage.getItem(this.storageKey)) || [];
-    }
+    private projects: APIProject[] = [];
```
CloudNiner: Can this be removed?

src/app/lab/lab.component.ts
```diff
@@ -24,22 +24,32 @@ import * as _ from 'lodash';
...
-    public project: Project;
+    public project: APIProject;
 
     constructor(private projectService: ProjectService,
                 private route: ActivatedRoute,
                 private router: Router) {
     }
 
+    // Save
+    @HostListener('window:beforeunload', ['$event'])
```
CloudNiner: Way better way to handle this case

src/app/services/project.service.ts
```diff
@@ -1,45 +1,43 @@
...
 
-    add(project: Project) {
-        this.projects.push(project);
-        this.save();
-    }
+    constructor(private apiHttp: ApiHttp) {}
 
-    get(id: string) {
-        return this.projects.find(project => project.id === id);
+    add(project: APIProject): Observable<APIProject> {
```
CloudNiner: Bah. These methods should have always returned Observables to make this local storage -> API change easier.

fungjj92:
Updated ~

PR #118, opened 2016-11-01
Fix chart duplication from bad merge

fungjj92:
## Overview

Fixes a bad merge error of PR 115 into PR 116 that slipped by.


fungjj92:
Trivial, merging

PR #121, opened 2016-12-30
Feature/fix lab

fungjj92:
Newest changes to CC API, particularly time_aggregations, broke the CC lab graphs. This should restore any lost functionality. You'll see some comments with temporary simplifications.

### DIFF
--- a/src/app/charts/chart.component.ts
+++ b/src/app/charts/chart.component.ts
@@ -51,7 +51,10 @@ export class ChartComponent implements OnChanges {
             indicator: this.chart.indicator,
             scenario: this.scenario,
             city: this.city,
-            climateModels: this.models
+            climateModels: this.models,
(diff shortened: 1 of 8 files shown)


src/app/charts/chart.component.ts
```diff
@@ -51,7 +51,10 @@ export class ChartComponent implements OnChanges {
             indicator: this.chart.indicator,
             scenario: this.scenario,
             city: this.city,
-            climateModels: this.models
+            climateModels: this.models,
+            // As a temporary solution, the time agg defaults to the 1st valid option.
```
rmartz: Do we have a github issue to add selecting non-yearly granularities? That'd be pretty handy to have

fungjj92: I don't think it has be discussed, but now that we do time aggs that seems like a necessary feature to lab.

PR #123, opened 2017-01-12
Download chart as image

flibbertigibbet:
## Overview

Add button to chart toolbar to download chart as an image (PNG).

### Demo

New toolbar button:
image

Clicking leftmost toolbar button downloads this:
extreme_cold_events

File name is the indicator name.

### Notes

The SVG to PNG conversion probably doesn't work on IE; this maybe could be accomplished by writing the SVG to canvas first.

Attempted to also offer option to share PNG on social media, but did not follow through, as it would be complicated by requiring a server-side component for authentication with the social media service, and could not be accomplished with standard social sharing widgets. Image would need to be uploaded first, then the returned identifier for the media object passed in to the status post. We would also have to put together our own edit/preview pane. I left in the intermediate commits for adding a social media icon popover to the chart toolbar button, in case that comes in handy later. Largely based on [this demo](http://codepen.io/labnol/pen/CenhB), with various CSS fixes.

For Twitter posting, [this endpoint](https://dev.twitter.com/rest/reference/post/statuses/update_with_media), though deprecated, would be the most straightforward to use for the post with image.

## Testing Instructions

 * Will need to `npm install` to pick up SVG-to-PNG converter package
 * Each chart should have new download toolbar button
 * Clicking image download button should immediately download appropriate chart, with CSS reflected


flibbertigibbet:
Extended file name. Ran into a curious issue with the typescript checker complaining about using dot notation on the city properties.

src/app/charts/chart.component.ts
```diff
@@ -70,7 +70,14 @@ export class ChartComponent implements OnChanges {
     }
 
     onDownloadImageClicked() {
-        this.socialService.downloadAsPNG(this.chart.indicator.name);
+        let fileName: string = [
```
CloudNiner: In ES6 prefer template literal syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

CloudNiner:
Yeah Object is probably too specific of a type for properties. If we're not going to explicitly list what's available on `city.properties`, it should be changed to `any` type. Same for `city.geometry`, but there are existing `@types` for geojson available we can use. Can be a separate issue to resolve.

flibbertigibbet:
Added GeoJSON typings and made `City` an alias for `Point`.

CloudNiner:
:+1:

PR #125, opened 2017-01-12
Feature/fix test runner

### LINKED ISSUE #124: Tests fail
`npm test` fails with:

```
Chromium 55.0.2883 (Ubuntu 0.0.0) ERROR
  Uncaught Error: Missing: SyncTestZoneSpec
  at /projects/climate-change-lab/config/spec-bundle.js:8043 <- webpack:///~/zone.js/dist/jasmine-patch.js:19:0
```

flibbertigibbet:
Updates to test runner. Fixes #124. `npm test` will now run without error, though the final output is inconsistent.

Now instead of returning an error on test run, debug output says our tests pass fine, but then npm reports that they failed. Perhaps this is because there are no tests.

Changes here are pulling in updates from AngularClass/angular2-webpack-starter. See angular/zone.js#407 for a discussion of the error issue.


flibbertigibbet:
Tried upgrading all test and some webpack packages, to no effect. Also pushed a commit to fix current build from clean package install.

flibbertigibbet:
My original suspicion was correct; just needed to set the flag `failOnEmptyTestSuite` to explicitly allow an empty test suite to pass.

CloudNiner:
Huzzah.

PR #127, opened 2017-01-17
Feature/update angular 2.4.3

fungjj92:
## Overview

Updating angular version to stay current but also to prepare for writing tests.

### Notes

Oddly smooth process...With the extra time, I adjusted the webpack config in development to allow for debugging. Also, the linter that runs per build should now check as it should have, but didn't. This should make developing way waaaayyy better.

## Testing Instructions

rm -rf node_modules
npm cache clear
npm i npm -g
npm start

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -24,7 +24,7 @@ module.exports = webpackMerge(commonConfig, {
 
   metadata: METADATA,
   debug: true,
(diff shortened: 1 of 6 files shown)


CloudNiner:
If providing test instructions, make sure you actually run through them to ensure they're correct. The actual list of instructions I used were:
- rm -rf node_modules
- npm cache clear
- npm install
- npm start

PR #128, opened 2017-01-20
Feature/add tests

flibbertigibbet:
## Overview

Fixes test and linter runners, adds some tests, and fixes linter warnings.

### Notes

Upgraded typescript to be able to upgrade linter package versions.
Switches back to Chrome karma launcher, as PhantomJS is non-functional, due to read-only error stack.
Due to the Chrome karma launcher, a browser window briefly flashes open during test run; apparently this cannot be disabled (see [this issue](karma-runner/karma-chrome-launcher#27)).

Code changes outside of configuration and added tests were all made to address linter warnings... there were a lot.

## Testing Instructions

 * `npm start` should run successfully, to output of 'OK' message (no typescript check issues)
 * `npm run lint` should exit successfully, with no warnings (if there are warnings, npm exits in error)
 * UI should still function as before
 * `npm test` should run seven tests successfully


config/spec-bundle.js
```diff
@@ -43,7 +43,7 @@ testing.TestBed.initTestEnvironment(
  * any file that ends with spec.js and get its path. By passing in true
  * we say do this recursively
  */
-var testContext = require.context('../src', true, /\.spec\.ts/);
+var testContext = require.context('../src/', true, /\.spec\.ts/);
```
CloudNiner: Huh. That's surprising.

config/webpack.common.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
   },
 
   resolve: {
-    extensions: ['', '.ts', '.js', '.json'],
+    extensions: ['','.ts', '.js', '.json'],
```
CloudNiner: Can this be reverted?

flibbertigibbet: Made that revert but failed to push it.

flibbertigibbet:
Test setup is only crazy for components, which requiring rendering. I think it may be related to our use of webpack; the official test examples in [the docs](https://angular.io/docs/ts/latest/guide/testing.html) are much simpler. Basically it proved necessary to import and setup every other component to test a single one of them, which defeats the intent to unit test them separately. So, compartmentalization of our code isn't the issue (more/smaller modules wouldn't help).

Hopefully the test frameworks and relevant documentation will mature to alleviate this.

Testing services, on the other hand, proved to be very straightforward, but our services at this point don't really do much other than interact with the API.

PR #134, opened 2017-02-27
Remove 'closes' keyword from PR template

flibbertigibbet:
To avoid Waffle auto-closing issues.


flibbertigibbet:
Trivial, merging. See azavea/climate-change-api#312

PR #135, opened 2017-03-21
Update DNS to climate.azavea.com

fungjj92:
## Overview

Switched from api.futurefeelslike.com --> climate.azavea.com

Connects https://github.com/azavea/climate-change-api/issues/337

### DIFF
--- a/example/constants.ts.example
+++ b/example/constants.ts.example
@@ -7,4 +7,4 @@
 export const defaultCity = {'id': 7, 'properties': {'name': 'Philadelphia','admin': 'PA'}};
 export const defaultScenario = 'RCP85';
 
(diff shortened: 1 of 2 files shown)


example/constants.ts.example
```diff
@@ -7,4 +7,4 @@
 export const defaultCity = {'id': 7, 'properties': {'name': 'Philadelphia','admin': 'PA'}};
 export const defaultScenario = 'RCP85';
 
-export const apiHost = 'https://api.staging.futurefeelslike.com';
+export const apiHost = 'https://staging.climate.azavea.com';
```
CloudNiner: Since people could theoretically check this project out, or fork it and make changes, and/or even just have access to the source code, let's update this to point to `climate.azavea.com` instead of staging since that's the production instance by default.

PR #147, opened 2017-06-14
Feature/awf/fix geojson types

CloudNiner:
## Overview

Fixes an error that I only saw on the Jenkins build host. Could not reproduce locally, but ssh'ed in and tested the fix there. A bit more detail in commit message.

Also switches jenkins build so that deploy only happens if build is successful.


CloudNiner:
Build succeeded and deploy was successful. Merging.

PR #156, opened 2017-06-23
Don't display Indicators with required params in sidebar

### LINKED ISSUE #141: Handle threshold indicators
They don't currently get graphed, since they take an extra parameter

Either handle graphing them, or filter them out of the indicator list.

I vote, for now, filter them out until or unless we decide to put effort into lab

CloudNiner:
## Overview

We need to make changes to the Lab to properly display any indicator with a required parameter. When a user selects one now, the graph just "loads" infinitely because we don't display an error message when the API request for data fails. Since allowing users to set values for required parameters would take significant design/dev effort, for now we just skip adding them to the sidebar.

Currently applies to:
- max temp threshold
- min temp threshold
- precip threshold

All other indicators with extra params, e.g. Cooling Degree Days, provide defaults for their extra params.

### Demo

Here's the new valid indicator list in the sidebar:
screen shot 2017-06-23 at 09 45 34

## Testing Instructions

 * Load up the lab. Shouldn't have any indicators that don't complete successfully (barring throttling errors)

Closes #141

### DIFF
--- a/src/app/sidebar/sidebar.component.ts
+++ b/src/app/sidebar/sidebar.component.ts
@@ -31,10 +31,16 @@ export class SidebarComponent implements OnInit {
     }
 
     private groupIndicators(indicators: Indicator[]) {
-        this.tempIndicators = indicators.filter(i => {
+        // Filter any indicators with "required" params, we can't currently display them
+        //  without changes to the Lab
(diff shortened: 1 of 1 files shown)


CloudNiner:
👍 I'll make that change.

PR #159, opened 2017-06-30
Feature/kak/fix test runner#145

### LINKED ISSUE #145: 'npm run test' fails in the VM
This will continue to be the case until we switch the test runner to a headless browser.

flibbertigibbet:
## Overview

Fixes test runner. Install Chromium in VM, so tests run via Chrome headless may be run within vagrant environment. Also adds a CI linter run option, for the upcoming Jenkins test/lint run task.

## Testing Instructions

 * `vagrant provision`
 * `vagrant ssh`
 * `cd /vagrant`
 * `npm run test`
 * Tests should run to completion

Fixes #145

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ node_modules/
 coverage/
 build/
 dist/
+violations.xml
 
 # Vagrant
 .vagrant/
--- a/ansible/roles/climate-change-lab.app/tasks/main.yml
+++ b/ansible/roles/climate-change-lab.app/tasks/main.yml
@@ -1,9 +1,10 @@
 ---
 
-  - name: Install build dependencies
+  - name: Install packages
     apt: name={{ item }} state=present
     with_items:
       - g++
+      - chromium-browser
 
   - name: "Install node dependencies"
     npm: path={{ app_dir }}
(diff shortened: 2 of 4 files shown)


fungjj92:
When was the Jenkins warnings plugin added? Can you point me to it?

flibbertigibbet:
I think we've had the warnings plugin set up pretty much since we had our own team Jenkins setup; it's used for other projects with linters. Not sure what you want to see. An example of the output chart? Can see one of those [here](http://urbanappsci.internal.azavea.com/job/azavea/job/climate-change-lab/view/change-requests/job/PR-160/).

There is a separate PR for the Jenkins configuration that needs this merged first.

flibbertigibbet:
If you just want to test the CI linter run, can run the `npm` command directly and check out the result in the `violations.xml` file (one of the formats the Jenkins linter supports).

fungjj92:
The description for commit 527111a made me think like it was a new thing. I think I just wanted to know more about outputting lint violations to file in the checkstyle format, a format I've not encountered before

flibbertigibbet:
Jenkins being in Java, it tends to like XML formatted things. Unfortunately the documentation for Jenkins isn't the best. This just happened to be a format `ESLint` would produce that Jenkins would consume. The regular linter run still produces human-readable output (in a non-weird format).

fungjj92:
OK thanks for explaining!

PR #160, opened 2017-06-30
Feature/kak/jenkins pipeline#151

### LINKED ISSUE #148: Run tests during CI build
Fail build if test failures.

Depends on #145

### LINKED ISSUE #151: Jenkins should only deploy to production lab site on changes to master branch
Currently deploys on all changes to develop.

flibbertigibbet:
## Overview

Switch to Jenkins pipeline to build project. Run tests and linter on Jenkins, reporting any linter warnings. For `develop` and `master` branch builds, also push site to S3, to staging or production site, respectively.

### Notes

Should remove the disabled non-pipeline Jenkins build configuration after this is merged.

## Testing Instructions

Since this is a Jenkins configuration PR, there's not really anything to do to test it locally. Can check out  the build logs [here](http://urbanappsci.internal.azavea.com/job/azavea/job/climate-change-lab/view/change-requests/job/PR-160/) and see the slack notifications produced in the channel.

Closes #151
Closes #148


flibbertigibbet:
build now

scripts/cibuild.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+vagrant up --provision
```
CloudNiner: While [not explicitly stated here](https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md), generally the only script that wraps its commands in `vagrant ssh -c` is the `./scripts/setup`. All other scripts are intended to be run within the VM. It would be good to refactor these bash scripts to follow that convention. The climate API project is still a good example to base a refactor on.

scripts/cibuild.sh
```diff
@@ -0,0 +1,17 @@
...
+vagrant ssh -c "
+    cd /vagrant
+    echo -e '\nRunning linter...'
+
+    # run linter with console output
+    npm run lint
```
CloudNiner: I don't think `./scripts/cibuild` is the appropriate place for running linting and tests. A more complete breakdown of what runs where might look like:
- ./scripts/setup
  - `vagrant up --provision`
- ./scripts/update
  - `npm install`
- ./scripts/test
  - lint
  - lint:ci
  - test
- ./scripts/cibuild
  - build:prod
- ./scripts/cipublish
  - `s3_website push`

scripts/cibuild.sh
```diff
@@ -0,0 +1,17 @@
...
+
+    # run linter with file output for Jenkins
+    npm run lint:ci
```
CloudNiner: It would be nice to conditionally run this command only when within the Jenkins environment. I believe there's a few ENV variables that only Jenkins sets that could be used to detect that state, but I can't seem to find the documentation for it. Maybe @azavea/operations can point us in the right direction?

scripts/cipublish.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+vagrant ssh -c "cd /vagrant
```
CloudNiner: Same comment here re: `vagrant ssh` wrapper

Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+          unstableTotalAll: '0',
+          usePreviousBuildAsReference: true
+        ])
+      }
+    }
+
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') {
+      // Publish to S3
+      stage('cipublish') {
+        // Decode the `AWS_ECR_ENDPOINT` credential stored within
```
CloudNiner: Stale comment

scripts/cibuild.sh
flibbertigibbet: This script is only intended to be run in the Jenkins environment (this is the `cibuild` script, only for building on the CI server.)

scripts/cibuild.sh
flibbertigibbet: Worth noting the `cibuild` script for the climate API project runs the `test` script, and therefore runs tests and linting, if indirectly. Since the scripting required to do so for this project is considerably less involved, I thought it made sense to consolidate.

Also note that provisioning runs `npm install`, so there is no need for a separate script for that.

scripts/cipublish.sh
flibbertigibbet: In that case, [Cicero usage](https://github.com/azavea/cicero/blob/develop/scripts/cibuild) needs to be updated as well.

scripts/cibuild.sh
tnation14: Chiming in to note that `scripts/cibuild` generally does call `scripts/test`. I normally prefer having all test content in `scripts/test` so that it can be run in development without also running any additional steps in `cibuild`.

scripts/cibuild.sh
tnation14: For your reference, [these](https://wiki.jenkins.io/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-belowJenkinsSetEnvironmentVariables) are the ENV variables set by Jenkins, but I think it's safe to assume that this script is only running in CI. One other note though: in [other projects](https://github.com/azavea/raster-foundry/blob/develop/scripts/cibuild#L24) we've included a conditional to only run a script if it's been invoked directly,  (i.e. not via `source`).

CloudNiner: True. The `npm run lint:ci` could still be run in `scripts/cibuild` even if `npm run lint` is moved into `scripts/test`. That would probably be a better way to handle it rather than conditionals.

scripts/cipublish.sh
CloudNiner: Indeed. Cicero doesn't look like a good up to date example -- it doesn't implement the full suite of scripts. It looks like that cibuild script was added quickly as a quick way to get the necessary functionality there.

scripts/cibuild.sh
flibbertigibbet: Why would the script be inadvertently executed via `source`?

scripts/cibuild.sh
CloudNiner: Sounds good re: npm install.

flibbertigibbet: That makes sense for projects that need it, but this is a single-project repository, which is why all this scripting seems excessive. Testing and building are both one-line commands, typically run outside the VM during development.

CloudNiner: This is an example of a trade off where we do a just a bit more work, but win in cross-project consistency where every developer here is able to immediately know where to go to run key project functions. And external developers are able to do the same because we document the scripts in the README. 

In either case, I'm fine with deferring this to #161.

scripts/cibuild.sh
tnation14: Although I don't think that the script is likely to be executed that way in this project, I just mentioned the convention because it's a good security measure for STRTA that may set variables that we don't want leaking into the CI environment.

Issue #173, opened 2017-07-26
Add historical data to lab chart


CloudNiner:
Should be included so that the default chart time range is 1950 to 2100.

The historical data should be requested at the same time as the future scenario requested by the user and combined with it to create one data series to pass to the chart component.

This task should also remove the "historical" entry from the Scenario dropdown in the top bar.

PR #186, opened 2017-08-01
Feature/awf/scenario toggle#174

### LINKED ISSUE #174: Scenario dropdown in top bar should be switched to a toggle
screen shot 2017-07-26 at 12 40 32

CloudNiner:
## Overview

Replaces the scenario dropdown in the top bar above the lab chart with a toggle, which only displays RCP45 and RCP85. Historical was removed with the intention to display it as part of the chart in #173.

### Demo

screen shot 2017-08-01 at 09 47 01
screen shot 2017-08-01 at 09 46 55

### Notes

Styling is still iffy, will need a quick pass by a designer, along with quite a few of these other issues at some point. I will open a separate issue to track follow-up designer tasks.

I began a refactor where in parent component/modules we only import from the `index.ts` file, rather than by the explicit file path to a component. This allows us to add/remove/update/move components and filenames without having to go through the tedious process of updating file paths all over the codebase. In addition, it makes it easy to see at a glance what the "public" interface of a module is by looking at that module's `index.ts` file. We should continue this practice in as many places as possible as we refactor the UI for this sprint.

## Testing Instructions

 * View a lab chart and the project add/edit page. You should see the dropdown replaced with a toggle in both cases. When the toggle is changed, the chart/project should update and request the other scenario.

Closes #174


flibbertigibbet:
Looks like the karma imports need updating to fix tests.

src/app/lab/components/scenario-toggle.component.ts
```diff
@@ -49,7 +44,7 @@ export class ScenarioDropdownComponent implements OnInit {
 
     private getScenarios() {
         this.scenarioService.list().subscribe(data => {
-            this.scenarios = data;
+            this.scenarios = data.filter(s => s.name === 'RCP85' || s.name === 'RCP45')
```
fungjj92: In the interest of constants book-keeping, consider saving constants 'RCP85' and 'RCP45' to a list and doing a `.includes(s.name)` or other check

CloudNiner:
Updated. I fixed up the following:
- The styling on the add/edit page
- Using includes and a declarative array rather than in place ORs for choosing the scenarios to include in the toggle
- A bug where clicking one of the toggle buttons on the add/edit page would submit the parent form

fungjj92:
** When the build failure is figured out

PR #189, opened 2017-08-01
Feature/kak/switch to json export#180

### LINKED ISSUE #180: Implement JSON download option
User should be able to download the full JSON API response used to construct the indicator chart

### LINKED ISSUE #188: Image export broken
Exporting an image PNG results in a black chart.

image

flibbertigibbet:
## Overview

Switch from CSV chart data export to JSON export, removing CSV export. Also fixes chart export image issue in #188. Refactored a couple of service names that were no longer accurate.

## Testing Instructions

 * Click export button above chart
 * Should get chart data as JSON file

Closes #180 
Fixes #188


src/app/services/data-export.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import 'rxjs/Rx';
+import * as _ from 'lodash';
```
fungjj92: lodash no longer used here

src/app/charts/chart.component.ts
```diff
@@ -62,7 +62,7 @@ export class ChartComponent implements OnChanges {
     }
 
     onExportClicked() {
-        this.csvService.downloadAsCSV(this.chartData);
+        this.dataExportService.downloadAsJSON(this.chartData);
```
fungjj92: `chartData` has undergone some transformation since the raw response from the API request to format for drawing the line graphs. It is similar, but different. Since my understanding is we are trying to mirror the API I think we might want the raw response data.

https://github.com/azavea/climate-change-lab/blob/develop/src/app/charts/chart.component.ts#L57

Thoughts @CloudNiner ?

flibbertigibbet: Yes, we want the raw response.

CloudNiner: Yeah, raw response, and `chartData` isn't that. We may need to save the most recent API indicator data response in local storage or memory so that it can be pulled here. 

Note some changes to the API -> indicator data -> chartData processing pipeline in #191 due to the addition of historical data to the chart.

PR #190, opened 2017-08-01
Feature/kak/move download buttons#179

### LINKED ISSUE #179: Move download button
New download button should contain options for:
- JSON used to generate chart
- PNG image of current chart state

Can remove existing CSV download option.

flibbertigibbet:
## Overview

Move the two chart download buttons into a drop-up below chart.

### Demo

image

### Notes

Our usage of `ngx-bootstrap` differs somewhat from the documented dropdowns/dropups [here](http://valor-software.com/ngx-bootstrap/#/dropdowns). Notably, this project defines its own bootstrap component styles. It was necessary to add some missing CSS to support drop up (get the menu to position above the button instead of below).

## Testing Instructions

 * Load a chart
 * Should have new dropdown below chart that toggles up
 * Download options in chart should fire download actions (note #180 changes these actions)

Closes #179

### DIFF
--- a/src/app/charts/chart.component.html
+++ b/src/app/charts/chart.component.html
@@ -63,18 +63,18 @@
 </div>
 
 <div dropdown [dropup]="true" class="btn-group">
-  <button dropdownToggle class="button dropdown-toggle" type="button"
-    id="exportDropUp" data-toggle="dropdown" aria-haspopup="true"
-    aria-expanded="true">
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
Okay. The added HTML was based on the dropdown template in `scenario-dropdown.component.ts`, which has two spaces.

PR #191, opened 2017-08-01
Include historical scenario in all indicator charts

### LINKED ISSUE #173: Add historical data to lab chart

CloudNiner:
## Overview

This PR adds an HTTP request to pull historical data for each indicator we chart in addition to the future data. For now the data is mashed together during the HTTP calls and in the `convertChartData` method.

### Demo

screen shot 2017-08-01 at 14 53 58

### Notes

I explored splitting the chart data used in LineGraphComponent into two separate lines to be drawn, one for historical data and one for future data, so that we could do things like draw or style the lines differently. However, that requires additional work and digging into the depths of the LineGraphComponent, so I deferred that for now until we decide the feature is warranted.

I also explored making a request to the scenario endpoint to get the actual Scenario object, rather than use a static, dummy one. In the end, all we care about is having scenario.name, so I left it as is. Otherwise we'd need significant changes to the logic of the ChartComponent to handle the additional async request, that doesn't gain us anything since we still need to look for the magic scenario.name to match against once we've pulled the list.

## Testing Instructions

- Load up a few different indicators, see historical values.

Closes #173

### DIFF
--- a/src/app/charts/chart.component.ts
+++ b/src/app/charts/chart.component.ts
@@ -1,5 +1,7 @@
 import { Component, EventEmitter, OnChanges, Input, Output, HostListener } from '@angular/core';
 
+import { Observable } from 'rxjs/Observable';
+
 import { Chart } from '../models/chart.model';
 import { ChartData } from '../models/chart-data.model';
 import { City } from '../models/city.model';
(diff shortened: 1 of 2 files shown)


CloudNiner:
Updated. The forkJoin operation returns an array of the results of each of the original observables in the order which they were added. It's effectively the same as the old `$q.all()` method.

We didn't think about this before, but for now I think it makes sense to just make the "future" data downloadable, and we can address if we want the historical data to be available for download as well later.

CloudNiner:
Opened #194 to address the historical/future data download discrepancy.

PR #192, opened 2017-08-02
Feature/jf/topbar changes

### LINKED ISSUE #172: Add units dropdown to bar above chart
screen shot 2017-07-26 at 12 33 25

Should include an option for each available unit that the indicator supports, list can be taken from `/api/indicator/` result from sidebar.

fungjj92:
## Overview

Adds the unit dropdown to the top bar and the settings modal.

### Demo

With no chart:

screen shot 2017-08-02 at 5 20 13 pm
screen shot 2017-08-02 at 5 20 24 pm

With chart:
screen shot 2017-08-02 at 5 20 07 pm
screen shot 2017-08-02 at 5 20 37 pm

### Notes

I am not entirely satisfied with this solution. I am thinking implementing an Observable would be a good idea, but that may also be overkill.

Also, the branch name is inaccurate.

## Testing Instructions

Check out https://github.com/azavea/climate-change-api/pull/608 in CC API and point your lab to the locally running API

Closes #172

### DIFF
--- a/src/app/lab/components/units-dropdown.component.ts
+++ b/src/app/lab/components/units-dropdown.component.ts
@@ -1,15 +1,13 @@
-import { Component, OnInit, Input } from '@angular/core';
+import { Component, Input } from '@angular/core';
 
-import { Indicator } from '../../models/indicator.model';
-import { ProjectData } from '../../models/project-data.model';
+import { Chart } from '../../models/chart.model';
 
 /*  Units Dropdown Component
 
     -- Requires project input
     Expected use:
         <ccl-units-dropdown
-            [projectData]="your_project.project_data"
-            [units]="avail_units">
+            [chart]="your_chart">
 */
 
 @Component({
@@ -19,31 +17,24 @@ import { ProjectData } from '../../models/project-data.model';
                 id="unitsDropdown" data-toggle="dropdown" aria-haspopup="true"
                 aria-expanded="true">
                 <i class="icon-flash"></i>
-                {{ projectData.unit }}
+                {{ chart.unit }}
                 <i class="icon-angle-down caret"></i>
               </button>
               <ul *dropdownMenu class="dropdown-menu" aria-labelledby="unitsDropdown">
-                <li *ngFor="let unit of units">
+                <li *ngFor="let unit of chart.indicator.available_units">
                   <a (click)="onUnitSelected(unit)"
                     placement="bottom">{{ unit }}</a>
                 </li>
               </ul>
             </div>`
 })
-export class UnitsDropdownComponent implements OnInit {
+export class UnitsDropdownComponent {
 
-    @Input() projectData: ProjectData;
-    @Input() units: [string];
-    public indicator: Indicator;
+    @Input() chart: Chart;
 
     constructor() {}
 
-    ngOnInit() {
-        this.indicator = this.projectData.charts[0].indicator;
-        this.projectData.unit = this.indicator.default_units;
-    }
-
     public onUnitSelected(unit: string){
-        this.projectData.unit = unit;
+        this.chart.unit = unit;
     }
 }
--- a/src/app/lab/components/units-dropdown.component.ts
+++ b/src/app/lab/components/units-dropdown.component.ts
@@ -1,0 +1,49 @@
+import { Component, OnInit, Input } from '@angular/core';
+
+import { Indicator } from '../../models/indicator.model';
+import { ProjectData } from '../../models/project-data.model';
+
+/*  Units Dropdown Component
+
+    -- Requires project input
+    Expected use:
+        <ccl-units-dropdown
+            [projectData]="your_project.project_data"
+            [units]="avail_units">
+*/
+
+@Component({
+  selector: 'ccl-units-dropdown',
+  template: `<div dropdown class="dropdown dropdown-units">
+              <button dropdownToggle class="button dropdown-toggle" type="button"
+                id="unitsDropdown" data-toggle="dropdown" aria-haspopup="true"
+                aria-expanded="true">
+                <i class="icon-flash"></i>
+                {{ projectData.unit }}
+                <i class="icon-angle-down caret"></i>
+              </button>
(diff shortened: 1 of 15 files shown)


src/app/lab/components/units-dropdown.component.ts
```diff
@@ -0,0 +1,40 @@
...
+                <li *ngFor="let unit of chart.indicator.available_units">
+                  <a (click)="onUnitSelected(unit)"
+                    placement="bottom">{{ unit }}</a>
+                </li>
+              </ul>
+            </div>`
+})
+export class UnitsDropdownComponent {
+
+    @Input() chart: Chart;
```
CloudNiner: This seems like a good example of where we could refactor such that this component only takes the information it needs. Instead of sending the whole chart into the component, we could just have the inputs be something like `unitOptions: string[]` and it could emit the unit selected via an `EventEmitter` that the chart component could listen for and then update the API call from there.

src/app/lab/components/units-dropdown.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    constructor() {}
```
CloudNiner: Is it necessary to define the constructor if it doesn't do anything?

src/app/lab/lab.component.ts
```diff
@@ -61,16 +69,25 @@ export class LabComponent implements OnInit, OnDestroy {
         }
     }
 
+    public resetChart() {
```
CloudNiner: What problem is the new `resetChart()` machinery in this class solving?

src/app/lab/components/units-dropdown.component.ts
CloudNiner: The problem with the solution as implemented is that there's no way for the units dropdown component to tell its parent explicitly that a new unit value was selected -- `chart.unit` isn't 'watched'. My guess is that this is the reason for the extra `resetChart()` machinery below.

Happy to pair today to attempt a refactor if that would be helpful!

fungjj92: Good suggestions, I'm on it!

src/app/lab/components/units-dropdown.component.ts
fungjj92: Good question, I don't know. I thought I had read something long ago that said yes, but upon searching now I'm not finding that or any consistency across open source Angular2 projects. Perhaps a convention we define team-wide, then. I'm fine to ixnay it here!

CloudNiner: I poked at the component section of the [Angular style guide](https://angular.io/guide/styleguide#components) and didn't see a preference explicitly mentioned, but every example omits the constructor when blank. I would generally agree -- if it's not required that it be implemented then it should be omitted if its not doing anything.

This is probably a more meta discussion, but in general we should defer to the Angular style guide for consistency across projects.

fungjj92:
Updated!

src/app/lab/components/units-dropdown.component.ts
```diff
@@ -0,0 +1,44 @@
...
+            (unitSelected)="onUnitSelected($event)">
+*/
+
+@Component({
+  selector: 'ccl-units-dropdown',
+  template: `<div dropdown class="dropdown dropdown-units">
+              <button dropdownToggle class="button dropdown-toggle" type="button"
+                id="unitsDropdown" data-toggle="dropdown" aria-haspopup="true"
+                aria-expanded="true">
+                <i class="icon-flash"></i>
```
CloudNiner: I'm not sure we want the lightning bolt here? The new wireframes don't have any icons in these controls at all.

src/app/lab/components/units-dropdown.component.ts
```diff
@@ -0,0 +1,44 @@
...
+                {{ unit || indicator.default_units }}
```
rmartz: Would this make sense to add to the API's response, so the lab doesn't need to know ahead of time what units it might encounter?

CloudNiner: Are you referring to including the human-readable names in the API response?

rmartz: Yeah, unless we already do that in which case don't mind me

CloudNiner: We don't send human-readable, just the short versions, e.g. "F", "C", "kg/m^2".

src/app/lab/components/units-dropdown.component.ts
```diff
@@ -0,0 +1,44 @@
...
+                  <a (click)="onUnitSelected(unit)"
+                    placement="bottom">{{ unit }}</a>
+                </li>
+              </ul>
+            </div>`
+})
+
+export class UnitsDropdownComponent {
+
+    @Input() indicator: Indicator;
```
fungjj92: I had thought I'd needed `indicator.default_unit` in addition to `indicator.available_units,` so i sent through indicator. Turns out no, just `indicator.available_units` because I set `chart.unit` upon chart initialization.

src/app/lab/components/units-dropdown.component.ts
fungjj92: oo hehe that slipped through

src/app/lab/components/units-dropdown.component.ts
fungjj92: I dunno. I think anyone looking at climate data through lab or the API would know what units they wanted. And for the ones they don't know, (basically, only K or kg/m^2) they probably don't want that anyway. Lastly how would you write kg/m^2 in a human readable way that is actually _more clear_ lol

fungjj92: If this were a really public tool like a NYT article graphic, human readable would make perfect sense. I don't think this is that audience

CloudNiner: Fair enough. We can leave as is and add later if user testing indicates its helpful!

PR #193, opened 2017-08-03
Add legend to chart with icons and description

### LINKED ISSUE #177: Add legend to indicator chart
In screenshot below, green line is average of selected models

In screenshot below, grey box/area on chart is the range between min/max of the selected models

screen shot 2017-07-26 at 12 48 43

sharph:
## Overview

This PR adds a legend to indicator charts. The unit is taken from the API response, as is the indicator description. `icon-minus` is used as the key for the line and the unicode character for "large black square" is used as the key for the range area.

### Demo

climatelegend-scrnshot

### Notes

I wasn't sure if there was a better way to represent the line and the square. Using the icon for "minus" in particular seems to be a bit of a hack, but it seems to match the line weight in the chart. The range area appears pretty light on my screen, so I darkened the color a little bit for the legend. Sometimes the unit could be represented better than the API represents it. For example "degrees F" might be better than "F." As for positioning, there wasn't a lot of context in the screenshot in the issue so I'm not sure how well this matches with what the designers intended.

## Testing Instructions

 * Click different charts
 * See changing description and units

Closes #177

### DIFF
--- a/src/assets/sass/components/_chart.scss
+++ b/src/assets/sass/components/_chart.scss
@@ -30,6 +30,30 @@ ccl-chart {
     padding-top: 1.5rem;
   }
 
+  .chart-legend {
+    padding: 2rem;
+    padding-top: 0;
+    color: #757575;
+    font-size: 11px;
+
+    .chart-legend-line {
+      color: #1E8CEC;
+    }
+
+    .chart-legend-icons {
+       padding-bottom: 0.5rem;
+    }
+
+    .chart-legend-minmax {
+      color: darken(#EFF4F7, 10%);
+
+      &:before {
+        padding-left: 0.5em;
+        content: '\2B1B'; // unicode large filled square
+      }
+    }
+  }
+
   .chart-options {
     border-top: 1px solid $gray-2;
     margin: 0;
--- a/src/assets/sass/components/_chart.scss
+++ b/src/assets/sass/components/_chart.scss
@@ -37,7 +37,7 @@ ccl-chart {
     font-size: 11px;
 
     .chart-legend-line {
-      color: #1E8CEC;
+      color: $chart-line;
     }
 
     .chart-legend-icons {
@@ -45,7 +45,7 @@ ccl-chart {
     }
 
     .chart-legend-minmax {
-        background-color: darken(#eef4f7, 10%);
+        background-color: darken($chart-minmax, 10%);
         width: 1rem;
         height: 1rem;
         display: inline-block;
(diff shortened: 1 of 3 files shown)


src/assets/sass/components/_chart.scss
```diff
@@ -30,6 +30,30 @@ ccl-chart {
...
+    .chart-legend-icons {
+       padding-bottom: 0.5rem;
+    }
+
+    .chart-legend-minmax {
+      color: darken(#EFF4F7, 10%);
+
+      &:before {
+        padding-left: 0.5em;
+        content: '\2B1B'; // unicode large filled square
```
CloudNiner: Ugh. I was hoping this would work, but no go on Chrome:

screen shot 2017-08-03 at 10 01 58

Red == the actual position of the span (by setting `background-color`)

screen shot 2017-08-03 at 10 01 49

It looks like Chrome isn't aware of this unicode character? "Black medium square" `\25FC` works fine, so does "upwards double arrow" `\21D1`.

An alternative is to give the `.chart-legend-minmax` a fixed width/height and set the background color rather than color. That looks something like:
```css
.chart-legend-minmax {
    background-color: #cddde6;
    width: 1rem;
    height: 1rem;
    display: inline-block;
    margin-left: 0.5rem;
}
```

Maybe @jfrankl has an idea as to why that specific unicode character doesn't work? Maybe its not a part of Chrome's default font family?

sharph: I think it's coming out in Chrome as a 3D shaded emoji black square. I'll probably use your alternative.

sharph: Fixed in https://github.com/azavea/climate-change-lab/pull/193/commits/760b431c3865a0de0c0b92015bd56209b1707cd8. Works fine in Firefox. I can't test in Chrome due to problems proxying through my laptop at the office.

src/assets/sass/components/_chart.scss
```diff
@@ -30,6 +30,29 @@ ccl-chart {
     padding-top: 1.5rem;
   }
 
+  .chart-legend {
```
CloudNiner: Can any of the colors in this newly added CSS block use the colors defined in `_variables.scss` or be tweaked slightly to use those colors or variations of them?

sharph: Addressed (and merge conflicts cleared) in https://github.com/azavea/climate-change-lab/pull/193/commits/111edca6fb32af70582b2be672297f61918dbccd

PR #195, opened 2017-08-03
Feature/kak/get api call button#181

### LINKED ISSUE #181: Implement "Get API Call" button + dialog
Dialog should contain:
- text box that shows the `curl` command that could be used to request the data used for the current chart state, and should update as the user changes options
- a button to the right of the text box that the user can click to copy the contents of the text box to their system clipboard

screen shot 2017-07-26 at 12 54 59

flibbertigibbet:
## Overview

Adds a popup for user to copy curl command to get chart data.

### Demo

image

### Notes

Needs a fontello icon for clipboard copy. Existing project icons don't seem to have anything appropriate.

## Testing Instructions

 * Click to copy curl command below chart
 * Should get confirmation tooltip that it copied
 * Should be able to run curl command and get data

Closes #181


src/app/charts/chart.component.ts
```diff
@@ -71,8 +74,16 @@ export class ChartComponent implements OnChanges {
             historical,
             future
         ).subscribe(data => {
+            const chartQuery = data[1].url;
+            delete data[1].url; // apart from URL, returned data is raw query response
             this.rawChartData = data[1];
-            this.chartData = this.chartService.convertChartData(data)
+            this.chartData = this.chartService.convertChartData(data);
+
+            this.curlCommand = ['curl -i "',
```
CloudNiner: Prefer native [ES6 template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

src/app/services/indicator.service.ts
```diff
@@ -10,15 +10,16 @@ import { apiHost } from '../constants';
...
  */
 @Injectable()
 export class IndicatorService {
     constructor(private apiHttp: ApiHttp) {}
 
     public getData(options: IndicatorQueryOpts) {
-        const url = `${apiHost}/api/climate-data/${options.city.id}/${options.scenario.name}` +
-                  `/indicator/${options.indicator.name}/`;
+
+        const getDetailUrl = `${apiHost}/api/climate-data/${options.city.id}/${options.scenario.name}` +
```
CloudNiner: Any particular reason for this change? Maybe `dataUrl` instead, since this isn't so much a 'detail' view?

flibbertigibbet: Reason why has been rebased away. Was going to maybe put the URL build in a separate, non-async method but decided it wasn't worth it. Sure, can rename back.

flibbertigibbet:
Pushed some updates. Gave up on trying to fix bootstrap styles for the input group button. Adding the `input-groups` bootstrap stylesheet should make the input field and button flush with rounded corners, but instead broke rendering of the button, so I gave up on trying to make the input group button work here.

image

PR #197, opened 2017-08-08
Feature/jf/time slider 2

### LINKED ISSUE #176: Implement time slider
Time slider defaults to full year range 1950-2100.

To start, only update the chart when the user stops dragging on one of the two ticks of the slider.
screen shot 2017-07-26 at 12 46 56

fungjj92:
## Overview

Ho ho! It's arrived. The time slider. This was the only updated and appropriate time slider lib out there. Luckily it was pretty straightforward to use despite the virtually non-existent documentation.

See: https://github.com/tb/ng2-nouislider

### Demo

screen shot 2017-08-08 at 4 38 57 pm

### Notes

Styling is obviously not on point, but I think it looks in a place to leave fine tuning to design.

## Testing Instructions

Checkout https://github.com/azavea/climate-change-api/pull/613/
Point your `constants.ts` at your locally running instance of CC API

`npm install` because of new packages
`npm run serve`
http://localhost:4200

Closes #176


src/app/charts/chart.component.ts
```diff
@@ -79,15 +100,22 @@ export class ChartComponent implements OnChanges {
...
 
             this.curlCommand = `curl -i "${chartQuery}" -H "Authorization: Token ` +
                                `${this.authService.getToken()}"`;
         });
     }
 
-    onSettingsToggleClicked() {
-        this.chart.showSettings = !this.chart.showSettings;
+    sliceChartData() {
+        this.chartData = _.cloneDeep(this.processedData); // to trigger change detection
```
fungjj92: This seems frivolous, but the latter filtering doesn't trigger change detection I think because a property rather than the whole object at `this.chartData` is being changed.

flibbertigibbet:
Could you cherry-pick the commit from #613 on here and close the other PR? Both easier to review together, and removal of the settings components in HTML is happening over on this PR already.

src/app/charts/chart.component.ts
```diff
@@ -79,15 +100,22 @@ export class ChartComponent implements OnChanges {
...
+        this.chartData[0]['data'] = _.filter(this.chartData[0]['data'], obj => {
```
CloudNiner: Lodash filter is unnecessary here, `Array.prototype.filter()` has full support and will be transpiled/polyfilled correctly by the build process:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

src/app/charts/chart.component.ts
```diff
@@ -79,15 +100,22 @@ export class ChartComponent implements OnChanges {
...
+            const year = obj['date'].getFullYear()
+            if(year >= this.dateRange[0] && year <= this.dateRange[1]) {
+                return obj
```
CloudNiner: This works, but not in the way you expect. filter expects a boolean response, so obj is true and the `undefined` returned in the default case is false. Can change this whole if block to:
```
return year >= this.dateRange[0] && year <= this.dateRange[1]
```

fungjj92:
@flibbertigibbet The other PR #613 is on on the CC API project

fungjj92:
Hm these Jenkins errors don't make sense:
`Error: Template parse errors:`
`Can't bind to 'config' since it isn't a known property of 'nouislider'.`

But `config` totally comes from nouislider:
https://github.com/tb/ng2-nouislider/blob/master/src/nouislider.ts#L69
Demos from the creator:
http://tb.github.io/ng2-nouislider/

CloudNiner:
Can you reproduce the same error on your machine?

flibbertigibbet:
Adding `NouisliderModule` to the `TestBed` imports in `login.component.spec.ts` will fix up the test failure.

src/app/charts/chart.component.ts
```diff
@@ -69,6 +89,7 @@ export class ChartComponent implements OnChanges {
             // Really, this should a user selectable option
             time_aggregation: this.chart.indicator.valid_aggregations[0]
         };
+        this.dateRange = [1950, 2100]; // reset time slider range
```
flibbertigibbet: Might be good to move these years to named constants. Do these relate to specific dates in the API project? If so, might want to note here and over there that if one set of dates gets changed, the other should be, too.

fungjj92:
When we discussed the feature we decided that we wanted to only update the chart when the user lets go of the slider. I think it would be possible though, same as we do the scrubber, with a constant listener.

fungjj92:
Nah I was waiting to dialogue with Kathryn about the question she posed. We're all good now!

Issue #182, opened 2017-07-26
Implement component and templates for indicator-specific parameters


CloudNiner:
An example of what this looks like:

![screen shot 2017-07-26 at 12 57 36](https://user-images.githubusercontent.com/1818302/28633503-05a73e80-7202-11e7-8988-6633bf3a0ba5.png)

Things to consider:
- Some indicators have parameters without defaults. For the lab, we can either choose arbtrary defaults or delay indicator chart load until a value is entered for these parameters
- Indicators have variable numbers of custom parameters
- A template will need to be written and added to the project for each indicator with extra parameters

CloudNiner:
Work with @jcahail and @fungjj92 or @CloudNiner to develop the templates for each indicator that needs one.

flibbertigibbet:
Regarding the example image: for the cooling degree days parameter, there is no `threshold_comparator` parameter, so there's no option for which "less than" is a choice; as I understand it, the results will always be for temperatures less than the `basetemp` parameter.

There are a number of other parameters not reflected in the example. In particular, I don't think `agg`, `custom_time_agg`, `years`, or `time_aggregation` are user-controllable elsewhere, but it may not make sense to throw them all in this new component. Do we have some more information on what, exactly, we'd like to control for each indicator?

CloudNiner:
Yes, design was unaware of the specifics for the Cooling Degree Days parameter, for a more helpful, complete example, change the title "Cooling Degree Days" in the example image to "Min Temperature Threshold". That indicator should align with the extra "indicator-specific" parameters available on the `min_temperature_threshold` indicator.

These templates would not include any parameters that are globally available on all indicators. This effectively means ignoring including in the template any parameter that is defined as part of [IndicatorParams](https://github.com/azavea/climate-change-api/blob/develop/django/climate_change_api/indicators/params.py#L131).

An easy way to get started might be to develop a template for each IndicatorParams subclass as those are the parameters that are "extra" and only apply to certain indicators.

flibbertigibbet:
Okay. Regarding `time_aggregation`, that looks to be globally available but with options dependent on the indicator. Currently there's a TODO in `chart.components.ts` to make it user-selectable (currently defaults to first option). Should that be part of this component?

flibbertigibbet:
Mapping of indicator parameters to the indicators that use them, for those that are not used by all indicators, based on results of `/api/indicators` self-describing endpoint:

```
[...]

PR #201, opened 2017-08-09
Feature/kak/indicator params cleanup

flibbertigibbet:
## Overview

Assorted small tidying made while investigating #182.

 - make `IndicatorParameter` a modeled type
 - move the query parameters (as opposed to path parameters) to a nested object

## Testing

 - shouldn't break anything :woman_shrugging:

### DIFF
--- a/src/app/models/indicator-parameter.model.ts
+++ b/src/app/models/indicator-parameter.model.ts
@@ -1,0 +1,6 @@
+export class IndicatorParameter {
+    name: string;
+    description: string;
+    required: boolean;
+    default?: any;
+}
--- a/src/app/models/indicator-query-opts.model.ts
+++ b/src/app/models/indicator-query-opts.model.ts
@@ -7,8 +7,10 @@ export interface IndicatorQueryOpts {
     indicator: Indicator;
     city: City;
     scenario: Scenario;
-    climateModels?: ClimateModel[];
-    years?: string[];
-    time_aggregation?: string;
-    unit?: string;
+    params: {
+        climateModels?: ClimateModel[];
+        years?: string[];
+        time_aggregation?: string;
+        unit?: string;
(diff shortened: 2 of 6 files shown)


src/app/models/indicator-query-opts.model.ts
```diff
@@ -7,8 +7,10 @@ export interface IndicatorQueryOpts {
     indicator: Indicator;
     city: City;
     scenario: Scenario;
-    climateModels?: ClimateModel[];
-    years?: string[];
-    time_aggregation?: string;
-    unit?: string;
+    params: {
```
fungjj92: Hm these nested in params diverges from the API response: 

screen shot 2017-08-09 at 5 10 29 pm

Maybe you were looking at indicator.parameters ?

fungjj92:
Adding the model for IndicatorParam is good though as we'll expect to use that data

src/app/models/indicator-query-opts.model.ts
flibbertigibbet: There are the query parameters passed to make the request and not intended to map to the response. Intention here is to distinguish them from the *path* parameters, which are the other things on this object.

fungjj92: Oh I see now. Sounds good!

src/app/models/indicator-parameter.model.ts
```diff
@@ -0,0 +1,8 @@
+/* tslint:disable:variable-name */
```
CloudNiner: Do any of these params actually fail the variable name check? It's probably better to only disable checks when necessary, right?

flibbertigibbet: No actually, was just defined on several other models.

Issue #209, opened 2017-08-17
Store extra parameters on project


flibbertigibbet:
If the selected chart has extra query parameters set in the user controls, save them on the project (adding fields). Also read the settings back when loading the chart.

See #182.

PR #210, opened 2017-08-17
Feature/kak/threshold parameter controls#206

### LINKED ISSUE #206: Implement controls for threshold parameters
See #182.

Implement user controls for modifying `threshold`, `threshold_units`, and `threshold_comparator`, for the indicators that use them.

flibbertigibbet:
## Overview

Adds controls above chart for user modification of extra query parameters used by the three threshold indicators. See #206 and #182 for discussion.

### Demo

image

### Notes

Uses Angular [reactive forms](https://angular.io/guide/reactive-forms#reactive-forms), which does not allow for using a Bootstrap dropdown, as it generates its own Material Design styled `select` component. Updated styles to try to match, but the area around the drop-down caret looks slightly different.

The options for `threshold_units` and `theshold_comparator` are not programatically discoverable in the indicator description, as are other options, such as `units`. Currently the options and defaults are coded into the threshold parameters component. These could maybe be moved out to a shared constants file for indicators with extra parameters in implementing the next set.

There seem to be some strange behavior around removing charts and re-adding them that may warrant opening a new issue.

Created #209 for saving the new control values to the project.

## Testing Instructions

 * Select an indicator from the sidebar whose name ends with 'Threshold' (there are three)
 * Should get controls above chart
 * Changing the controls should update the chart as expected

Closes #206.


src/app/charts/extra-components/index.ts
```diff
@@ -0,0 +1 @@
+export * from './threshold.component';
```
CloudNiner: It might be nice for this directory to have a more specific name, maybe `extra-params-components` which describes a bit more about the types of components we would expect to see in this directory. `extra-components` is too vague. Can I use this directory for all components that are used within chart component, or is intended just as a folder for the components that provide the "extra indicator params templates"?

src/app/charts/chart.component.html
```diff
@@ -11,6 +11,12 @@ <h2 class="chart-label">
         </button>
     </div>
 </div>
+<div *ngIf="chart.indicator.thresholdIndicator" class="chart-options">
```
CloudNiner: How do you envision adding additional sets of extra params components here? Just as a long list of if blocks? I see adding additional extra params components getting unwieldy fairly quickly. There are a limited number to add though, so some significantly abstracted intermediate component might be overkill as well.

src/app/models/indicator.model.ts
```diff
@@ -10,6 +10,7 @@ export class Indicator {
     available_units?: string[];
     default_units?: string;
     parameters?: IndicatorParameter[];
+    thresholdIndicator: boolean;
```
CloudNiner: I don't think this is an extensible approach. The Indicator class (which should actually be an interface) is intended to match the Indicator object returned from the API, and now it contains additional items that describe an instance's state, rather than its properties. It also requires extra logic in the `list()` method of the Indicator service that needs to be duplicated anwhere an Indicator is returned from the API, such as in Indicator.getData, which returns an indicator as one of the top level objects in its response.

Since this is a "derived" property of an Indicator, meaning that its state can be calculated from another property of indicator, it makes more sense as a method on the Indicator class, or in an indicator utils service that has a method such as `isThresholdIndicator(indicator: Indicator) => boolean`. The first approach is more OOP-like, but has the same disadvantage as the solution here in that it requires `new`ing each indicator object from all the API responses so that you get instances of the class rather than just an object that matches the Indicator interface.

We haven't done enough work with TypeScript/Angular to have a good sense of how to approach these types of problems unfortunately, so I don't have a sense of which is best. But I think either are preferable to having significant logic required to properly create these Indicator instances that exists outside of the Indicator model file.

Another option for the method that tells what "type" of indicator a given indicator is, rather than having a boolean method for each type, might be something like `getType(indicator: Indicator) => IndicatorTypeEnum`.

src/app/services/indicator.service.ts
```diff
@@ -24,6 +25,16 @@ export class IndicatorService {
         // Generate query params
         const searchParams: URLSearchParams = new URLSearchParams();
         const optParams = options.params;
+
+        // append extra parameters for threshold indicators
+        if (options.indicator.thresholdIndicator) {
+            // abort request if chart is in flux (these parameters are required)
+            if (!optParams.threshold) return Observable.of({url: ''});
```
CloudNiner: Given the extra checks that TypeScript gives us, I think this is a dangerous thing to return. While the method technically returns an `Observable<any>`, really its expected to return something like `Observable<IndicatorDataResponse>`, which is a large convoluted type that we haven't actually defined anywhere. `Observable.of({url: ''})` wouldn't match that nebulous type that we haven't defined for sure. It's probably better to raise a specific error with `Observable.throw()` and have the calling code handle this specific problem.

src/app/models/indicator-query-opts.model.ts
```diff
@@ -12,5 +12,10 @@ export interface IndicatorQueryOpts {
         years?: string[];
         time_aggregation?: string;
         unit?: string;
+
+        // threshold indicator parameters
```
CloudNiner: TypeScript interfaces can be extended. These might be better added as a separate
```
interface ThresholdIndicatorQueryOpts extends IndicatorQueryOpts {
    threshold: Number;
    threshold_units: string;
    threshold_comparator: string;
}
```
Allows us to explicitly send the proper parameter type with a given indicator and better describe the type (e.g. for ThresholdIndicatorQueryOps, these three extra params are not optional.

src/app/charts/chart.component.ts
```diff
@@ -137,6 +158,11 @@ export class ChartComponent implements OnChanges {
         this.imageExportService.downloadAsPNG(this.chart.indicator.name, fileName);
     }
 
+    public onThresholdSelected($event) {
+        delete $event.event; // everything else is a threshold parameter
```
CloudNiner: This comment seems like something that could be explicitly addressed by TypeScript by defining an interface for the event emitter that sends this object, rather than having to implicitly 'know' that everything other than `$event.event` is what we want, which is both opaque and an easy way to introduce bugs when refactoring or adding features later.

src/app/charts/chart.component.ts
```diff
@@ -76,24 +77,44 @@ export class ChartComponent implements OnChanges {
...
+    }
+
+    updateChart(extraParams: any) {
         this.chartData = [];
         this.rawChartData = [];
+
+        let params = {
+            climateModels: this.models,
+            unit: this.unit || this.chart.indicator.default_units,
+            // As a temporary solution, the time agg defaults to the 1st valid option.
```
CloudNiner: Hmm. We haven't addressed this. For now we should always be sending `time_aggregation: 'yearly'`, although its pretty likely that index 0 will always be 'yearly'. Definitely out of scope for this task, worth being a separate issue to address this since I think this happens in a few places.  We could also improve our handling of these by using the new string enums introduced in TypeScript 2.4 so that we wouldn't have to litter magic strings like 'yearly' throughout our code.

src/app/services/indicator.service.ts
CloudNiner: Is this check in there because of the behavior observed in #211?

src/app/charts/chart.component.html
flibbertigibbet: There are five varieties of components we need, or maybe just three, if the overlapping ones can be combined. Is that too many?

src/app/models/indicator.model.ts
flibbertigibbet: Was thinking myself of moving this as well as some other parameter-specific configuration to a shared constants file for all the extra parameters, mentioned in the notes.

src/app/models/indicator-query-opts.model.ts
flibbertigibbet: :+1: went looking for info on that but didn't immediately come up with any

src/app/charts/chart.component.ts
flibbertigibbet: Okay. This was the recommended approach from what I'd found, but it is unwieldy to have to go hang extra arguments onto an existing object. I think the thing passed does need to be of the special `event` type though.

src/app/charts/chart.component.html
CloudNiner: Not sure. We can leave as is and evaluate that in one of the later PRs that adds the additional components.

src/app/charts/chart.component.ts
CloudNiner: Gotcha. Do you have a link to that approach? If it's standard convention, then I'd be more ok with it. 

Maybe we could split the difference and do something like:
```es6
public onThresholdSelected($event) {
    // arbitrarily chose data as a more generic name that isn't part of the Event spec
    const thresholdParams = $event.data as ThresholdIndicatorQueryOpts
    ...
}
```
Feels a bit more clear, and still gives easy access to both the data packaged with the event and the event itself. Thoughts?

src/app/charts/chart.component.ts
flibbertigibbet: Created #212

src/app/charts/chart.component.ts
flibbertigibbet: Probably not a standard convention, just something I'd found I think from a blog post and one or two StackOverflow discussions, though I seem to have closed those tabs already. I'll try fiddling with it to come up with something better.

src/app/models/indicator.model.ts
flibbertigibbet: Went with a utility module, since it's just checking string membership in arrays.

Although it is a derived property, for performance reasons, it would be more efficient to cache the results somewhere, and the indicator object would seem to be an obvious place for that. Efficiency might not be an issue, though.

flibbertigibbet:
Addressed everything, I think, and changed @fungjj92 to reviewer since @CloudNiner will be out.

src/app/charts/chart.component.ts
```diff
@@ -76,24 +81,46 @@ export class ChartComponent implements OnChanges {
...
+            unit: this.unit || this.chart.indicator.default_units,
+            // TODO: #212
+            // As a temporary solution, the time agg defaults to the 1st valid option.
+            // Really, this should a user selectable option
+            time_aggregation: this.chart.indicator.valid_aggregations[0]
+        }
+
+        if (isThresholdIndicator(this.chart.indicator.name)) {
+            params = _.extend(params, extraParams);
+            this.isThresholdIndicator = true;
```
fungjj92: Which resolves this unwanted element persistence:

fungjj92: Hm actually, this only resolves switching between threshold and non-threshold indicators. I still get unwanted persistence this way when switching between threshold indicators..

src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    thresholdForm: FormGroup;
+
+    @Input() comparators: any[] = [
+        {'key': 'gte', 'label': 'greater than or equal to'},
+        {'key': 'lte', 'label': 'less than or equal to'},
+        {'key': 'gt', 'label': 'greater than'},
+        {'key': 'lt', 'label': 'less than'}
+    ];
+    @Input() thresholdUnits: any[] = [
+        {'key': 'K', 'label': 'Kelvin'},
```
fungjj92: Perhaps save temp and precip key/label object lists to global variables

fungjj92:
Since class vs interface came up a bit:
https://jameshenry.blog/typescript-classes-vs-interfaces/

src/app/charts/chart.component.ts
flibbertigibbet: Actually it is (should be) a function now and not a property. Fixed.

src/app/charts/extra-params-components/threshold.component.ts
flibbertigibbet: Moved out to separate class variables, could relocate if we need them elsewhere.

flibbertigibbet:
Modified to reset form on indicator change, as Angular won't do that for you; that had been masked by a fix for a related issue with events being ignored on initialization, since sorted out with the added lifecycle methods. This should be ready for another look.

PR #215, opened 2017-08-22
Update style to match new mockups

jfrankl:
## Overview

Update style to match latest mockups.

### Demo

image

### Notes

A developer is needed to make the chart work at full height.


fungjj92:
Lookin spiffy!

I pushed up a commit for line graph height.

Did you mean to get rid of the remove chart button? If yes, that will require some dev follow-up to remove associated logic.

fungjj92:
@flibbertigibbet  Pulling in Kathryn because I wrote some of this code, and I can't rightfully review my own code :P

fungjj92:
Styling all looks good to me @jfrankl

PR #222, opened 2017-08-25
Add historical curl command

### LINKED ISSUE #194: Update JSON export /  API calls button to include historical request
Now that the scenario selector only allows the choice of the future scenarios, and historical data is always included in the chart, the "Download the data for this chart" operation and the API `curl` call no longer match all of the data in the chart, since the chart includes both historical and future data, and the JSON download/`curl` components only make the future data available. 

We should refactor the JSON download/`curl` components in some way such that the user understands that the two datasets require two different queries, and make them both available for `curl` and download.

fungjj92:
## Overview

We want to provide historical and future curl commands since we provide both datasets to the user

### Demo

screen shot 2017-08-25 at 4 25 00 pm

## Testing Instructions

Check that both work as intended

Closes #194

### DIFF
--- a/src/app/charts/chart.component.html
+++ b/src/app/charts/chart.component.html
@@ -69,19 +69,16 @@
                             <div class="drowdown-item">
                                 <h4>Get API Call</h4>
                                 <p>Use our API to access these data regularly for your own research or application.</p>
-                                <div class="input-group">
-                                    <input type="text" class="form-control curl-command"
-                                        [(ngModel)]="curlCommand"
-                                        placeholder="curl command for API query"
-                                        #curlCommandInput>
-                                    <button type="button" class="input-group-btn"
-                                        tooltip="Copied!"
-                                        triggers="" #copiedPopup="bs-tooltip"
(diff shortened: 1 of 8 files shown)


src/app/charts/chart.component.html
```diff
@@ -65,15 +65,30 @@ <h2 class="chart-label">
                                 <h4>Get API Call</h4>
                                 <p>Use our API to access these data regularly for your own research or application.</p>
                                 <div class="input-group">
+                                    <span class="curl-label">Historical data</span>
```
CloudNiner: It would be nice to see this `div.input-group` DRYed into a separate `CopyCurlComponent` or something of that nature. There are clear inputs/outputs here with a DOM/functionality that is almost exactly the same -- only the value of the input and its placeholder changes between the two.

fungjj92:
Good suggestion. I thought it might be too heavy handed, but it turns out I like it after all.

fungjj92:
YAHOOO Jenkins passed. I always forget to add new components to the test spec.

PR #224, opened 2017-08-30
Ensure proper data loads for every chart display request

### LINKED ISSUE #211: Changing indicator during data load displays old response
Attempt to view and indicator that loads slowly (can replicate by throttling network speed in chrome dev tools). While the indicator is load, click another indicator. When the first request finishes, it will populate the chart view even though we've switched to a new indicator.

Should be able to fix by cancelling the http observable subscription if its not complete when we handle the indicator click event.

CloudNiner:
## Overview

A user could click on a new indicator before the current one
finished loading, leading to the old data displaying with
the new indicator selected. This prevents that by canceling
old data requests when we destroy the ChartComponent or
start loading new data.

### Demo

screencapture mov

## Testing Instructions

Testing is a bit odd, because once you've made a request for every indicator, the cached results make indicator display fast. Which is generally good, but not if we want to test this. So, before starting your npm server, add the line:
```js
searchParams.append('noCache', 'true');
```
to line 55 in `indicator.service.ts` then reload the app.

Once you've done that, load some indicator other than "Accumulated Freezing Degree Days" and "Average High Temperature". Then quickly click "Accumumlated Freezing Degree Days", then "Average High Temperature" like in the demo gif. You should only ever see the data for "Average High Temperature" display. Only the latest indicator clicked should ever display. This can be verified by clicking a few indicators in the sidebar before the first one clicked loads. 

Closes #211

### DIFF
--- a/src/app/charts/chart.component.ts
+++ b/src/app/charts/chart.component.ts
@@ -1,7 +1,17 @@
-import { AfterViewInit, Component, ChangeDetectorRef, EventEmitter, OnChanges, Input, Output,
-    HostListener } from '@angular/core';
+import {
+    AfterViewInit,
+    ChangeDetectorRef,
+    Component,
+    EventEmitter,
+    HostListener,
+    Input,
+    OnChanges,
+    OnDestroy,
+    Output
+} from '@angular/core';
 
 import { Observable } from 'rxjs/Observable';
+import { Subscription } from 'rxjs/Subscription';
 
 import { Chart } from '../models/chart.model';
 import { ChartData } from '../models/chart-data.model';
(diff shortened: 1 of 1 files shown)


src/app/charts/chart.component.ts
```diff
@@ -1,7 +1,17 @@
...
+    AfterViewInit,
+    ChangeDetectorRef,
+    Component,
+    EventEmitter,
+    HostListener,
+    Input,
+    OnChanges,
+    OnDestroy,
+    Output
+} from '@angular/core';
```
fungjj92: Yay for this formatting

Issue #217, opened 2017-08-24
Navbar changes incompatible with threshold indicators


fungjj92:
Changes to the nav bar options (model, scenario etc) don't propagate threshold indicator options in the request because they are captured from threshold option change events, and aren't saved on the chart. The produces an incomplete request and infinite chart load.

See line: https://github.com/azavea/climate-change-lab/pull/210/files#diff-d713ba72d21e2173ab63fddb34fd01e9R114

jcahail:
May be fixed by #209.

PR #225, opened 2017-08-30
Feature/jf/save extra proj params

### LINKED ISSUE #213: Refactor chart removal
Chart removal is a little messy. Several chart-related settings have been added and are force- removed/reset alongside charts. If possible, refactor charts such that all its associated settings are contained on the chart so removal will reset all associated settings

Run into especially with #210

fungjj92:
## Overview

Lots of things!
There were some loose-ends from adding threshold params that caused strange behavior. This PR pulls in a few issues that tied together to complete the feature.

### Notes

I left comments in the code to explain some of the decisions I made to work with Angular's lifecycle hooks (https://angular.io/guide/lifecycle-hooks#lifecycle-hooks), which became a little tricky with parent-child communication and reactive forms. Happy to talk through anything via messages or in person and/or provide more robust code comments!

## Testing Instructions

Checkout CC API https://github.com/azavea/climate-change-api/pull/649 and run locally
Set your apiHost to localhost
`npm run serve`

Make sure all threshold indicators work when you:
- select them
- select another one
- select non-thresh indicators
- change thresh params
- save the project and come back (aka leave the view)
- change navbar settings

Closes #213, #209, #217

### DIFF
--- a/src/app/charts/extra-params-components/threshold.component.ts
+++ b/src/app/charts/extra-params-components/threshold.component.ts
@@ -73,7 +73,6 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
         // Since valueChanges triggers initially before parent is ready, wait until
         // parent is ready here and trigger it to draw chart with extra parameters.
         this.thresholdParamSelected.emit({data: {
-            'event': null,
             'threshold_comparator': this.thresholdForm.controls.comparatorCtl.value,
             'threshold': this.thresholdForm.controls.thresholdCtl.value,
             'threshold_units': this.thresholdForm.controls.thresholdUnitCtl.value
@@ -82,6 +81,7 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
 
     ngOnChanges(changes: any) {
         // listen for the indicator label to be set before changing options and defaults, if needed
+
         if (this.label.indexOf('Precipitation') > -1) {
             this.thresholdUnits = this.thresholdPrecipitationUnits;
             this.thresholdForm.reset({
--- a/src/app/charts/extra-params-components/threshold.component.ts
+++ b/src/app/charts/extra-params-components/threshold.component.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, EventEmitter, OnChanges, Input, Output, OnInit } from '@angular/core';
+import { AfterViewInit, Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 
 import { Indicator } from '../../models/indicator.model';
@@ -13,7 +13,7 @@ import * as _ from 'lodash';
   selector: 'ccl-threshold-parameters',
   templateUrl: './threshold.component.html'
 })
-export class ThresholdComponent implements AfterViewInit, OnChanges, OnInit {
+export class ThresholdComponent implements AfterViewInit, OnInit {
 
     @Input() indicator: Indicator;
     @Input() extraParams: any;
@@ -53,20 +53,6 @@ export class ThresholdComponent implements AfterViewInit, OnChanges, OnInit {
     @Output() thresholdParamSelected = new EventEmitter<any>();
 
     constructor(private formBuilder: FormBuilder) {}
-
-    ngOnChanges(changes: any) {
-        /* Ignore change detection:
-            - before initalization
-            - from extraParams @Input whose initial values are all we want
-        */
-        if (changes.indicator && this.thresholdForm) {
-            this.thresholdForm.reset({
-                thresholdUnitCtl: this.defaultUnit,
-                comparatorCtl: this.defaultComparator,
-                thresholdCtl: this.defaultThreshold
-            });
-        }
-    }
 
     ngOnInit() {
         // must create form on init instead of constructor to capture @Input values
--- a/src/app/charts/extra-params-components/threshold.component.ts
+++ b/src/app/charts/extra-params-components/threshold.component.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, EventEmitter, OnChanges, Input, Output } from '@angular/core';
+import { AfterViewInit, Component, EventEmitter, OnChanges, Input, Output, OnInit } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 
 import * as _ from 'lodash';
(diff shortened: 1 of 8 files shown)


src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -48,54 +50,62 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
...
-                'threshold': form.thresholdCtl,
-                'threshold_units': form.thresholdUnitCtl
-            }});
-        });
+    ngOnChanges(changes: any) {
+        /* Ignore change detection:
+            - before initalization
+            - from extraParams @Input whose initial values are all we want
+        */
+        if (changes.label && this.thresholdForm) {
```
CloudNiner: Looked at this a bit more as promised. I think this is as good as it gets. Unfortunately, I'm still very concerned how extra logic like this in these template components will scale as we add additional ones, but that's a problem that's above the scale of this PR.

The only other alternative would be to switch the two explicit `@Input()` and `@Output` decorators that deal with `extraParams` to a single two way binding via `[()]`. But, this doesn't allow you to manually limit when we propagate changes up to the parent component. So it's no good. The `@Input` decorator doesn't have any additional options for ignoring changes after the first one, and in fact, that's not what you want here. You want the form to reset anytime the `label` property changes, which indirectly means the indicator this component is referring to, changes. 

Maybe all we do here is cleanup the comment to describe more the behavior we want, rather than just restating what the code is doing. e.g. switch the comment to something like "Only reset the form and push changes to the parent when the indicator changes via the label property, and when the form exists".

src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -48,54 +50,62 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
 
     @Output() thresholdParamSelected = new EventEmitter<any>();
 
-    createForm() {
-        this.thresholdForm = this.fb.group({
-            comparatorCtl: [this.defaultComparator, Validators.required],
-            thresholdCtl: [this.defaultThreshold, Validators.required],
-            thresholdUnitCtl: [this.defaultTemperatureUnit, Validators.required]
-        });
+    constructor(private fb: FormBuilder) {}
```
CloudNiner: As noted when looking at this with you, change to `private formBuilder: FormBuilder`

src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -48,54 +50,62 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
...
+                comparatorCtl: this.defaultComparator,
+                thresholdCtl: this.defaultThreshold
+            });
+        }
     }
 
-    constructor(private fb: FormBuilder) {
-        this.createForm();
+    ngOnInit() {
+        this.createForm(); // must create form on init instead of constructor to capture @Input values
```
CloudNiner: Comment on its own line

src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -48,54 +50,62 @@ export class ThresholdComponent implements AfterViewInit, OnChanges {
...
+            this.thresholdParamSelected.emit({data: {
+                'event': event,
+                'threshold_comparator': form.comparatorCtl,
+                'threshold': form.thresholdCtl,
+                'threshold_units': form.thresholdUnitCtl
+            }});
+        });
+    }
+
+    private evaluateVariable() {
```
CloudNiner: Since we need to know the default unit, maybe we should pass `Indicator` instead of `label` as an input, such that we could just pull the default unit from there rather than attempting to magically derive it based on substrings in the label.

src/app/charts/line-graph.component.ts
```diff
@@ -79,7 +79,7 @@ export class LineGraphComponent implements OnChanges, AfterContentInit {
     }
 
     /* Executes on every @Input change */
-    ngOnChanges(): void {
+    ngOnChanges(changes?: any): void {
```
CloudNiner: Was this change necessary for some reason? I don't think we need to redeclare the type of this method.

src/app/lab/lab.component.ts
```diff
@@ -89,9 +93,13 @@ export class LabComponent implements OnInit, OnDestroy {
             return;
         }
         this.removeChart();
-        this.indicator = indicator;
-        const chart = new Chart({indicator: indicator,
-                                 unit: indicator.default_units});
-        this.project.project_data.charts = [chart];
+        // Trigger lifecycle evaluation to truly destroy the chart child component
```
CloudNiner: This comment should probably have a bit more description of _why_ we want to completely destroy this component when we select a different indicator, so the context is apparent for the next person that looks at this.

src/app/charts/line-graph.component.ts
fungjj92: Leftover from playing around, good catch

src/app/charts/extra-params-components/threshold.component.ts
fungjj92: Hm upon review,`Indicator.default_units` doesn't necessarily coordinate to pr/temp unit (C,F,K,in,etc). It can be `days`for some indicators, so derivation is necessary in some form. I updated it to depend on `Indicator.variables,` which I think is more readable.

src/app/charts/extra-params-components/threshold.component.ts
fungjj92: I just realized that we never need to reset the form to defaults, since we destroy the threshold component along with the chart component between indicator selections. I removed `ngOnChanges()` and everything works the same, and we avoid constant listening for `@Input` -- whose constant listening depends on having `ngOnChanges()`

fungjj92:
Updated

src/app/charts/extra-params-components/threshold.component.ts
CloudNiner: Nice catch.

src/app/lab/lab.component.ts
```diff
@@ -89,9 +93,15 @@ export class LabComponent implements OnInit, OnDestroy {
             return;
         }
         this.removeChart();
-        this.indicator = indicator;
-        const chart = new Chart({indicator: indicator,
-                                 unit: indicator.default_units});
-        this.project.project_data.charts = [chart];
+        /*  Trigger lifecycle to truly destroy the chart component & its children
```
CloudNiner: I still dislike the decision to destroy/recreate, as completely destroying components and rebuilding is heavy handed and leads to cascading troublesome design decisions, such as the latest comment about removing the ngOnChanges no longer being necessary because we destroy/create. If we were ever to refactor and remove this destroy/create cycle, then the child ThresholdComponent would be broken because it no longer watches for the `@Input indicator` changes. As we continue to build based on this implicit behavior, it becomes easier to make similar decisions in the future which would further compound this issue.

With that said, I understand the impetus for it and would like to see the remaining tasks unblocked. A way to simplify this may become clear as we work through those tasks.

Issue #212, opened 2017-08-19
Modify selection of time aggregation parameter


flibbertigibbet:
Resolve `TODO` in `chart.component.ts`.

Either make `time_aggregation` query parameter user-selectable, or if always yearly, move to an `enum` or other form of constant.

See commentary on #182 and #210.

PR #228, opened 2017-08-31
Feature/kak/update dependencies#187

### LINKED ISSUE #187: Update packages
Update dependencies

flibbertigibbet:
## Overview

Update npm dependencies, using `ncu -a`. Fixes some resulting server errors.

### Notes

Doing this task now in case the updates to the karma Chrome launcher might help in addressing some of the spurious Jenkins test run failures. If not, will try the `--no-sandbox` flag workaround [described here](https://github.com/karma-runner/karma-chrome-launcher/issues/73#issuecomment-236597429).

## Testing Instructions

 * Should still serve without errors and behave as expected

Closes #187.


src/app/models/threshold-indicator-query-opts.model.ts
```diff
@@ -1,9 +1,16 @@
+import { ClimateModel } from './climate-model.model';
 import { IndicatorQueryOpts } from './indicator-query-opts.model';
 
 export interface ThresholdIndicatorQueryOpts extends IndicatorQueryOpts {
     params: {
         threshold: Number;
         threshold_units: string;
         threshold_comparator: string;
+
+        // from base
```
CloudNiner: I originally did this as well after the TS2.4 upgrade, but then went further down the rabbit hole and refactored IndicatorQueryOpts such that IndicatorQueryOpts.params is now `interface IndicatorParams` instead of `any`, and can be extended by `ThresholdIndicatorParams`. `ThresholdIndicatorQueryOpts` disappears. 

No action necessary here, just more of a heads up that a larger change related to this is coming in the PR for #212. It enforces type checking far better in a few places, and avoids the awkwardness of having to redefine parameters as is done here.

PR #231, opened 2017-08-31
Feature/awf/cleanup time agg param#212

### LINKED ISSUE #212: Modify selection of time aggregation parameter

CloudNiner:
## Overview

Replaces the `string` type of the `time_aggregation` indicator param with a new enum type `TimeAggParam`. This allows for improved type checking.

Assigned both of you so you can see this change. @flibbertigibbet because I mentioned I was making it in a previous PR you authored, and @fungjj92 so it can inform your work on the percentile indicator params.

### Demo

No user facing changes

### Notes

To make this easier, I replaced the subclassing of IndicatoryQueryOpts by making IndicatorQueryOpts.params have a new IndicatorParams interface that can be extended instead of the QueryOpts interfaces. This provides a much improved type interface for these, and allows us to better type check the params as they're passed around the app, especially when we construct requests to IndicatorService and when we pass the params around via EventEmitters.

The models folder is getting pretty big, and its now very unwieldy to import one item from each file everywhere. I considered adding an index.ts to the models folder and exporting all of the items in that folder there and updating import statements across the app, but that would have cluttered the intent of this PR. Maybe a good separate issue.

## Testing Instructions

App should continue to function

Closes #212


CloudNiner:
Most of below lifted from an offline discussion, but copied here for visibility (@jcahail):

I chose to go with only the `yearly` time agg option and not provide a user selectable dropdown. The quarterly and monthly options provide too much noise unless a user is zoomed in to a few year's time range, at which point it's not really possible to glean useful information from it.

Monthly option for all years:
screen shot 2017-09-01 at 10 04 56

Monthly option at ~5yr range:
screen shot 2017-09-01 at 10 06 32

We could potentially revisit this if we hear a bunch of users requesting it, but for now its less effort and less complicated for users to just omit the option.

src/app/charts/extra-params-components/threshold.component.ts
```diff
@@ -62,11 +62,11 @@ export class ThresholdComponent implements AfterViewInit, OnInit {
     ngAfterViewInit() {
         // Since valueChanges triggers initially before parent is ready, wait until
         // parent is ready here and trigger it to draw chart with extra parameters.
-        this.thresholdParamSelected.emit({data: {
+        this.thresholdParamSelected.emit({
```
flibbertigibbet: Why is the `data` object removed here? You had asked for it to be added in another PR.

src/app/models/indicator-params.model.ts
```diff
@@ -0,0 +1,10 @@
+import { ClimateModel } from './climate-model.model';
+import { TimeAggParam } from './time-agg-param.enum';
+
+export interface IndicatorParams {
+    climateModels?: ClimateModel[];
```
flibbertigibbet: Just noticed these are all optional. Shouldn't they be required?

src/app/services/chart.service.ts
```diff
@@ -17,7 +17,6 @@ export class ChartService {
 
     private timeOptions = {
           'yearly': '%Y',
-          'daily': '%Y-%m-%d',
```
flibbertigibbet: Should there be a quarterly format added here?

src/app/charts/chart.component.ts
```diff
@@ -115,18 +116,15 @@ export class ChartComponent implements OnChanges, OnDestroy, AfterViewInit {
         this.cancelDataRequest();
     }
 
-    updateChart(extraParams: any) {
+    updateChart(extraParams: IndicatorParams) {
         this.cancelDataRequest();
         this.chartData = [];
         this.rawChartData = [];
```
fungjj92: I'm not understanding why the typecheck doesn't complain when ThresholdIndicatorParams are sent through

src/app/models/indicator-params.model.ts
```diff
@@ -0,0 +1,10 @@
+import { ClimateModel } from './climate-model.model';
```
fungjj92: Now there are two models:
indicator-parameters.model.ts
indicator-params.model.ts

in the models folder, and that is confusing

src/app/charts/chart.component.ts
CloudNiner: Because `interface ThresholdIndicatorParams` extends `interface `IndicatorParams`

src/app/services/chart.service.ts
CloudNiner: Hmm, we don't use it but it wouldn't hurt to be thorough

src/app/models/indicator-params.model.ts
CloudNiner: No, because none of these params are required to make an API request, they all have defaults.

src/app/charts/extra-params-components/threshold.component.ts
CloudNiner: I think I had misunderstood what the original problem was based on the comments before, combined with forgetting the details of how `EventEmitter`s work. What's here now is the proper way to return data via event emitters, if you do not need the [DOM Event](https://developer.mozilla.org/en-US/docs/Web/API/Event)

src/app/charts/chart.component.ts
fungjj92: Ah interesting. Extending works the opposite direction I was expecting, then. Cool.

src/app/models/indicator-params.model.ts
CloudNiner: Any suggestions for better names? otherwise I'll try to come up with something

fungjj92: We've been using the term "extra params"
Maybe 
indicator-extra-params?

fungjj92: Or perhaps indicator-query-parameters

src/app/services/chart.service.ts
CloudNiner: Actually, can't do this now. It would require additional refactoring. In the quarterly case, we need to convert the keys in the API (e.g. 2010-3) into the start date for a "quarter" which would be July 1, 2010 for 2010-3, and that can't be done with `D3.timeParse`. We'd need this refactor for pretty much all of the other remaining valid time aggs if we chose to display them in the lab UI.

src/app/models/indicator-params.model.ts
CloudNiner: We already have indicator-query-opts and these params aren't always the subset that are the extra params. They're really the GET params to an API request. What about the following refactor:
- `IndicatorQueryOpts -> IndicatorRequestOpts`
- `IndicatorParams -> IndicatorGetParams`
- Leave `IndicatorParam` unchanged

fungjj92: OK IndicatorRequestOpts
#2 -- > IndicatorParamsList? or IndicatorQueryStringParams?
OK IndicatorParam

CloudNiner: `IndicatorQueryParams` would just be a shorter version of your second option that isn't any less descriptive. How about that? Definitely like your second suggestion better than the first.

PR #246, opened 2017-09-06
Add controls for historic and percentile indicators

### LINKED ISSUE #205: Implement controls to modify both historic range and percentile
See #182.

Implement user controls for historic range and percentile, for indicators that accept both parameters (and not just one or the other).

flibbertigibbet:
## Overview

Add extra parameter controls for the two indicators that take both historic range and percentile as extra parameters.

Simply stacks the two controls on top of each other.

### Demo

image

### Notes

Would be prettier and less repetitive to create an additional control with different text and both controls on one line. Went with this for now to get something implemented quickly.

## Testing Instructions

 * View chart for either extreme heat or extreme precipitation
 * Check both controls work
 
Closes #205

### DIFF
--- a/src/app/charts/extra-params-components/extra-params.constants.ts
+++ b/src/app/charts/extra-params-components/extra-params.constants.ts
@@ -13,16 +13,19 @@ const basetempIndicatorNames = [
 
 const historicIndicatorNames = [
     'heat_wave_duration_index',
-    'heat_wave_incidents'
+    'heat_wave_incidents',
+    'extreme_heat_events',
+    'extreme_precipitation_events'
 ];
 
 const percentileIndicatorNames = [
     'percentile_high_temperature',
     'percentile_low_temperature',
-    'percentile_precipitation'
+    'percentile_precipitation',
+    'extreme_heat_events',
+    'extreme_precipitation_events'
 ];
 
-// TODO: concat additional extra parameter names to this array for #205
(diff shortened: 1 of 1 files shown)


CloudNiner:
I know this isn't a result of this PR, but now that I'm looking at it again, it feels particularly unfriendly. Can we adjust the historic base range year dropdown to default to the year that the API defaults to, rather than present a blank item? It's very much not helpful for the user to see a blank space, and it only hurts to communicate the intent of the control, especially if blank is the default, which it was for me:

screen shot 2017-09-06 at 16 34 46

I'm fine with that either as a part of this PR or as a separate one. @jcahail do you have thoughts on this?

Spacing between the two items also seems particularly large. If its trivial to do it would be nice to reduce that to something more reasonable. Hopefully that's not another browser issue. If it is can you open a new issue and give it the design tag?

flibbertigibbet:
Blank is the default because it's not a required parameter. Is there some text you'd like to see for that instead, like 'none' or something?

jcahail:
Let's make this a separate issue, but I'd agree that having the default be blank is not very helpful.

flibbertigibbet:
Since these are the only two optional parameters with overlap in indicators, messing with the spacing seems like it would be less productive than adding a new component type with modified text for them both on one line. I'll make an issue to do that.

flibbertigibbet:
(Fix for tests should be coming in separately; pass locally)

PR #260, opened 2017-09-19
Redirect to login on auth error response

### LINKED ISSUE #259: Handle 401 API response
If API returns a `401` response to an API request, clear stored credentials (log out) and redirect to login page.

Unlikely to happen in production. In development, if the API host for the project is changed without manually logging out and back in to the site, data simply fails to load and a console error gets logged:

image

See #254, which handled the more common case of credentials missing.

flibbertigibbet:
## Overview

If server returns 401 or 403 auth error, log out and redirect to log in again.

### Notes

Handling the error in each HTTP method is not necessary, as each of them call `request` in turn ([source here](https://github.com/angular/angular/blob/4.4.2/packages/http/src/http.ts#L42-L185)).

## Testing Instructions

 * Log in to the site
 * With the server still running, change the `apiHost` in `src/app/constants.ts`
 * When page live reloads, it should redirect to the login page with a 401 error in the console

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Fixes #259.

### DIFF
--- a/src/app/auth/api-http.service.ts
+++ b/src/app/auth/api-http.service.ts
@@ -20,7 +20,12 @@ export class ApiHttp extends Http {
     }
 
     request(url: string | Request, options?: RequestOptionsArgs): Observable<Response> {
-        return super.request(url, this.appendAPIHeaders(options));
+        return super.request(url, this.appendAPIHeaders(options)).catch((error: Response) => {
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Feel free to re-assign @fungjj92

PR #267, opened 2017-09-25
Feature/kak/dataset selector#258

### LINKED ISSUE #257: Model selector should only show models valid for the current dataset
Currently it shows all models from `/api/climate-model/` but should actually show the model list from the selected dataset using `/api/dataset/` instead:

screen shot 2017-09-18 at 15 19 36

### LINKED ISSUE #258: Add a dataset selector to the Lab interface
So a user can toggle between viewing LOCA or NEX-GDDP data.

### LINKED ISSUE #262: Update models
Some of the API models have changed in the API and should also be updated in the lab as well.

For example, `city` no longer has `map_cell` but now has a`datasets` property.

flibbertigibbet:
## Overview

Adds a `dataset` selector.

Sends `dataset` query parameter. Filters `models` list to those valid for given `dataset`. Invalid, selected models remain selected but its checkbox control is disabled in the modal. Disables `dataset` options invalid for the current city selection.

Depends on [companion PR for the project schema](azavea/climate-change-api#701).

### Demo

image

image

### Notes

If a change in the selected city causes the `dataset` for the project to become invalid, the `dataset` will be changed to the default (NEX-GDDP) if that is valid, or the first valid option if not.

Currently there is no 'LOCA' data available for the development instance. To fake the 'LOCA' dataset being valid for a city, you can create a `ClimateDataCityCell` without data in the API project database:

 * `vagrant ssh`
 * `./scripts/manage shell_plus`

Then in the Django shell:
```python
# get a city
city = City.objects.first()

# create a LOCA map cell for the city
new_cell = ClimateDataCityCell(city=city, map_cell=city.map_cell_set.first(), dataset=ClimateDataset.objects.get(name='LOCA'))

new_cell.save()

# should now have two map cells for city, one for each dataset
city.map_cell_set.all()
```

Note that since there will be no actual data loaded, selecting 'LOCA' in the Lab will result in a blank chart with some chart-related console errors (about `trailing garbage` and `toFixed`). But, this should be sufficient for testing UI component interactions.

## Testing Instructions

 * Check out the [companion API PR](https://github.com/azavea/climate-change-api/pull/701) and start its server
 * Edit `apiHost` in `src/app/constants.ts` to point to 'http://localhost:8080'
 * `yarn run serve`
 * Try [creating a new project](http://localhost:4200/#/new), checking fields cross-update as appropriate
 * Try modifying project fields (city/dataset/models) from charts page, checking updates

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Closes #258 
Closes #257
Closes #262

### DIFF
--- a/src/app/lab/components/dataset-toggle.component.ts
+++ b/src/app/lab/components/dataset-toggle.component.ts
@@ -20,6 +20,7 @@ import * as _ from 'lodash';
     <div class="dropdown button-group">
         <button class="button"
                *ngFor="let s of datasets"
+               [disabled]="!isValidDataset(s)"
                [ngClass]="{'active': s.name === projectData.dataset.name}"
                (click)="onDatasetClicked(s, $event)">
             {{ s.name }}
@@ -49,6 +50,11 @@ export class DatasetToggleComponent implements OnInit {
         if (event) {
             event.preventDefault();
         }
+    }
+
+    // helper that checks if a given dataset is available for the currently selected city
+    public isValidDataset(dataset: Dataset): boolean {
+        return _.includes(this.projectData.city.properties.datasets, dataset.name);
     }
 
     private getDatasets() {
--- a/src/app/lab/components/dataset-toggle.component.ts
+++ b/src/app/lab/components/dataset-toggle.component.ts
@@ -57,15 +57,11 @@ export class DatasetToggleComponent implements OnInit {
 
     // set to default, or first valid option for the selected city
     private selectDefaultDataset() {
-        let dataset = this.datasets.find((s) => {
-            return s.name === this.DEFAULT_DATASET_NAME;
-        });
+        let dataset = this.datasets.find(s => s.name === this.DEFAULT_DATASET_NAME);
 
         // if the standard default dataset is not valid, use first valid option
         if (!this.isValidDataset(dataset)) {
-            dataset = this.datasets.find((s) => {
-                return this.isValidDataset(s);
-            });
+            dataset = this.datasets.find(s => this.isValidDataset(s));
         }
 
         this.onDatasetSelected(dataset);
--- a/src/app/lab/components/dataset-toggle.component.ts
+++ b/src/app/lab/components/dataset-toggle.component.ts
@@ -20,7 +20,7 @@ import * as _ from 'lodash';
     <div class="dropdown button-group">
         <button class="button"
                *ngFor="let s of datasets"
-               [disabled]="!isValidDataset(s)"
+               [disabled]="!checkValidDataset(s)"
                [ngClass]="{'active': s.name === projectData.dataset.name}"
                (click)="onDatasetSelected(s, $event)">
             {{ s.name }}
@@ -42,6 +42,9 @@ export class DatasetToggleComponent implements OnInit {
     }
 
     public onDatasetSelected(dataset: Dataset, event?: Event) {
+        if (!this.isValidDataset(dataset)) {
+            this.selectDefaultDataset();
+        }
         this.projectData.dataset = dataset;
         this.projectData.models.forEach(model => {
             model.enabled = _.includes(dataset.models, model.name);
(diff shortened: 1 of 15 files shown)


fungjj92:
All unchecked models look the same in the model list, whether or not they're disabled. If the text could be greyed, or that line could be ~crossed out~, or something more obvious than it is now, the UX would be more friendly

flibbertigibbet:
Unselected, disabled models appear greyed out, as do the selected, disabled models.

fungjj92:
The checkbox yes. It's so subtle I can hardly see it haha. This definitely is helpful to me:

screen shot 2017-09-26 at 1 40 17 pm

src/app/lab/components/dataset-toggle.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        });
+
+        if (event) {
+            event.preventDefault();
+        }
+    }
+
+    // set to default, or first valid option for the selected city
+    private selectDefaultDataset() {
+        let dataset = this.datasets.find((s) => {
```
fungjj92: remove parens? (s) -> s

src/app/lab/components/dataset-toggle.component.ts
```diff
@@ -0,0 +1,104 @@
...
+            return s.name === this.DEFAULT_DATASET_NAME;
+        });
+
+        // if the standard default dataset is not valid, use first valid option
+        if (!this.isValidDataset(dataset)) {
+            dataset = this.datasets.find((s) => {
```
fungjj92: same

src/app/lab/components/model-modal.component.ts
```diff
@@ -46,10 +58,15 @@ export class ModelModalComponent implements OnInit {
     }
 
     public updateProjectClimateModels() {
+        this.disableClimateModels();
         this.projectData.models = this.filterSelectedClimateModels();
         this.updateButtonText();
```
fungjj92: updateButtonText() is broken because the check is for # models selected against the full list of models.

this.projectData.models.length === this.climateModels.filter(m=> m.datasets.includes(this.projectData.dataset.name)).length

src/app/models/climate-model.model.ts
```diff
@@ -2,5 +2,6 @@ export interface ClimateModel {
     name: string;
     description: string;
     base_time: string;
+    enabled?: boolean;
     selected?: boolean;
```
fungjj92: also add `datasets: string[];` :)

src/app/lab/components/dataset-toggle.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        if (event) {
+            event.preventDefault();
+        }
+    }
+
+    // set to default, or first valid option for the selected city
+    private selectDefaultDataset() {
+        let dataset = this.datasets.find(s => {
+            return s.name === this.DEFAULT_DATASET_NAME;
+        });
```
fungjj92: these short functions can be simplified:
` let dataset = this.datasets.find(s => s.name === this.DEFAULT_DATASET_NAME);`

src/app/lab/components/dataset-toggle.component.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        this.onDatasetSelected(dataset);
+    }
+
+    // check if a given dataset is valid and change it if not and it is selected
+    public checkValidDataset(dataset: Dataset): boolean {
+        const isValid = this.isValidDataset(dataset);
+        if (!isValid && this.projectData.dataset &&
+            this.projectData.dataset.name === dataset.name) {
+            // If this dataset is selected but not valid, change to a valid dataset.
```
fungjj92: Do you have a workflow example such that I can test this behavior?

flibbertigibbet: In the 'Notes' section on this PR, there are instructions for adding (faking) a 'LOCA' dataset for a city. You can use that city in the lab to test out switching between datasets.

Specifically for this case:
 - select the city in the lab for which LOCA is valid, and select LOCA
 - change to a city for which LOCA is not valid, using the city picker in the upper left
 - the dataset selector should change back to NEX-GDP and LOCA should be greyed out

src/app/lab/components/model-modal.component.ts
flibbertigibbet: Good find. Can just use the count of project `dataset` models instead.

src/app/models/climate-model.model.ts
flibbertigibbet: The property is not getting used currently and duplicates the information on the `dataset` object (list of valid models), but for completeness it does make sense to add it.

src/app/lab/components/dataset-toggle.component.ts
fungjj92: This works. I ran into an issue bigger than this of checking that the scenario/dataset/city combo yields data or not. That check seems out of scope.

flibbertigibbet:
Bump (ready for review)

src/app/lab/components/dataset-toggle.component.ts
flibbertigibbet: I believe that a no data situation should only have occurred in testing this when a faked dataset was selected (see notes section). So this shouldn't be an issue.

flibbertigibbet:
Created #268 for styling, as I think we might want to consider switching away from a modal control for the model picker.

PR #273, opened 2017-10-04
Always display units dropdown

### LINKED ISSUE #261: Unit selector does not appear until chart selected
After creating a new project but before selecting a chart, there is a header label in the top bar for 'UNITS', but no selector beneath it.

image

flibbertigibbet:
## Overview

If no chart has been selected yet for a project, display the header units selector, with a single "N/A" dummy option.

Before, selector label would display but not the component below it. Added 'N/A' option since an empty, blank selector looked strange, and users might not recognize immediately that it is tied to the selected  indicator options.

### Demo

image

## Testing Instructions

 * Create a new project
 * Before selecting a chart, should see dropdown under units label
 * No console errors

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Fixes #261


flibbertigibbet:
Good point. Updated to put default string in the component template instead, and default the value to an empty string (which would also fail on a request, but won't mask the bug if we fail to set the parameter properly).

PR #280, opened 2017-10-09
Feature/kak/update precip units#276

### LINKED ISSUE #276: Change precipitation threshold units to rates
In https://github.com/azavea/climate-change-api/issues/664 we changed the API to use precipitation as a rate for Precipitation Threshold as well as Percentile Precipitation. Percentile Precipitation is able to ingest the new changes dynamically, but Precipitation Threshold needs to be updated to reflect the new options for `comparator_units`. They can be one of:
- `kg/m^2/s`
- `in/day`
- `mm/day`

These better reflect the difference in scale (Previously all were raw units like `kg/m^2` and `in`, which caused them to all be invisibly handled internally as `(unit)/s`.

flibbertigibbet:
## Overview

Update query units used by precipitation threshold chart to match API, which now uses rates.

### Demo

precip_units

### Notes

Existing lab projects with a precipitation threshold chart selected will remain broken until the chart or query unit selection gets updated.

## Testing Instructions

 * Select precipitation threshold indicator from sidebar
 * Chart should load with any of the three query unit options

Closes #276

### DIFF
--- a/src/app/charts/extra-params-components/extra-params.constants.ts
+++ b/src/app/charts/extra-params-components/extra-params.constants.ts
@@ -37,9 +37,9 @@ export const TemperatureUnits: any[] = [
  ];
 
 export const PrecipitationUnits: any[] = [
-    {'key': 'mm', 'label': 'millimeters'},
-    {'key': 'in', 'label': 'inches'},
-    {'key': 'kg/m^2', 'label': 'kg/m^2'}
+    {'key': 'mm/day', 'label': 'millimeters per day'},
+    {'key': 'in/day', 'label': 'inches per day'},
+    {'key': 'kg/m^2/s', 'label': 'kg/m^2/s'}
 ];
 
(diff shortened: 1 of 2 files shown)


CloudNiner:
I don't think we can only change precip units to the rate versions. The total_precipitation indicator still uses the non-rate version of these. It's likely we'll need to add an extra category here to ensure the lab properly handles PrecipUnits and PrecipRateUnits as defined in: https://github.com/azavea/climate-change-api/blob/develop/django/climate_change_api/indicators/unit_converters.py#L78-L106

You can see which indicators use rates and totals by grepping for `PrecipRateUnitsMixin` and `PrecipUnitsMixin` here: https://github.com/azavea/climate-change-api/blob/develop/django/climate_change_api/indicators/indicators.py

flibbertigibbet:
Total precipitation still works:
image

It doesn't pass query units, since only the threshold parameters send extra parameters.

Issue #283, opened 2017-10-12
Refactor Lab chart components to a separate module


CloudNiner:
Use same repo created for #282 to host this new component module.

Goal of this task is to move the following components to a new module in an external repo:
- LineChartComponent
- ChartComponent
- All of the dropdown components

And refactor the Lab to properly use the new external dependency.

CloudNiner:
Blocked by #282

PR #292, opened 2017-10-30
Remove ProjectData dependencies from Lab Components

CloudNiner:
## Overview

Part one of a two part strategy for splitting the lab components off to climate-change-components. Required removing all references to project models from the components to be moved.

### Demo

None, app should continue to work as expected

### Notes

CityDropdownComponent stays in Lab for now. It has a local dependency on ng2-autocomplete which is complicated to import, because that has a dependency on the apiHttpProvider in Lab which configures the apiHost and token auth properly. Dug a bit into this rabbit hole and bailed out for now. This is probably fine as CityDropdown isn't one of the components we immediately need, and a location search typeahead will likely be built differently in temperate and not tied to Climate API cities directly.

Since I was refactoring, theres a commit tacked onto the end that refactors most of the use of lodash out of the project, with the exception of the use of cloneDeep. Also updates the lodash dep to only be the code for the cloneDeep method, to prevent temptation to use things like `_.map` in the future when `Array.map` is valid ES6, with polyfills as appropriate handled by the underlying build process.

## Testing Instructions

Checkout, update deps with `yarn` and make sure the app works the same as before.

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Connects #283 

@maurizi didn't assign you, but you might want to take a look as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "d3-tip": "^0.7.1",
     "file-saver": "^1.3.3",
     "jquery": "^3.2.1",
-    "lodash": "^4.17.4",
+    "lodash.clonedeep": "^4.5.0",
     "moment": "^2.18.1",
     "ng2-nouislider": "^1.6.2",
     "ngx-bootstrap": "^1.9.1",
--- a/src/app/charts/chart.component.ts
+++ b/src/app/charts/chart.component.ts
@@ -27,7 +27,7 @@ import { isBasetempIndicator,
          isPercentileIndicator,
          isThresholdIndicator } from 'climate-change-components';
 
-import * as _ from 'lodash';
+import * as cloneDeep from 'lodash.clonedeep';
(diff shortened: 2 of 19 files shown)


CloudNiner:
Special care should be taken in testing the ModelModal. I _think_ I properly changed the references to the different climate model arrays, and I didn't see any errors there, but it should be double checked.

fungjj92:
screen shot 2017-10-31 at 9 11 57 am

Broken on threshold indicators

fungjj92:
Looks really good! Honestly no addtl comments. I agree about City Dropdown since we're going to do single city import/ingests

CloudNiner:
The precip threshold error is a separate one, unrelated to the changes made here. Opened https://github.com/azavea/climate-change-components/issues/6

PR #303, opened 2017-11-03
Make sidebar sections collapsible

### LINKED ISSUE #300: Make "Temperature" and "Precipitation" headings on the left nav collapsible
Open by default.  The list is long and you have to scroll down to see Precipitation.

flibbertigibbet:
## Overview

Accordion sidebar temperature and precipitation section headers.
Give same hover styling as sub-elements (blue, with pointer cursor).

Open by default.

### Demo

After clicking on first header to collapse:
image

### Notes

Also minor version bump to `ngx-bootstrap` library, not necessitated by anything.

## Testing Instructions

 * `yarn install && yarn run serve`
 * Sidebar headers should change to blue with pointer cursor on hover to indicate clickable
 * Clicking sidebar header should expand/collapse section
 * Sections should both be expanded by default

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Closes #300.


fungjj92:
Seeing wacky behavior in firefox, but not chrome
screen shot 2017-11-06 at 12 17 42 pm
screen shot 2017-11-06 at 12 17 47 pm

flibbertigibbet:
Was able to reproduce with Firefox, but noticed it would resolve on hover over the chart.

Pushed a number of dependency updates. At first was just a couple of updates to resolve "unmet peer dependency" warnings, which then necessitated further updates to fix spurious Typescript errors, apparently related to [outstanding rxjs issues](ReactiveX/rxjs#2539).

flibbertigibbet:
Given this is a browser issue, could we make an issue for Firefox chart sizing as well? We have a number of other browser-specific issues outstanding which might be resolved with a shared fix.

flibbertigibbet:
Specifically this is likely related to the [line graph resize](https://github.com/azavea/climate-change-components/blob/develop/src/lib/modules/charts/components/line-graph/line-graph.component.ts#L274), which now lies in another repository.

fungjj92:
And thus arrives a downside to factoring out lab components into a separate repo :( It makes sense to me that this chart issue is fixed with this work because they are directly connected.

src/app/app.module.ts
```diff
@@ -8,6 +8,7 @@ import { HttpModule, XHRBackend, RequestOptions } from '@angular/http';
 // 3rd party modules
 import {
     BsDropdownModule,
+    CollapseModule,
```
fungjj92: can this instead just be pulled in in `sidebar.module.ts`?

flibbertigibbet: There is no `sidebar.module.ts`

fungjj92: d'oh i got confused, thinking we modularize this project like Temperate

flibbertigibbet:
This did not introduce the chart resize issue but surfaced an existing issue; go to staging and delete the `ccl-indicator-list` element from under the sidebar header, and it does the same thing. Chart resize should not be dependent on the sidebar height.

flibbertigibbet:
Created azavea/climate-change-components#13

PR #306, opened 2017-11-07
Remove login page from the Lab in favor of using the API login page.

maurizi:
## Overview

This relies on a new endpoint in the API to allow the Lab to request the API token of a currently logged in API user, using it to both determine if a user is authenticated as well as to obtain an API token for requests.

### Notes

I considered taking advantage of the API and the Lab both being hosted on the same domain by simply setting the token on a cookie when the user logs into the API, but decided against it as we would have needed a different strategy for development anyway.

## Testing Instructions
 * Check out https://github.com/azavea/climate-change-api/pull/755 and make sure your Lab environment is set up to use your local API server.
 * Attempting to go to the Lab unauthenticated should redirect to the API login page
 * Logging into the API should leave the user at the API homepage
 * Attempting back to the Lab should now work without redirecting
 * Logging out from the Lab should redirect the user to the API logout page
 * Attempting to go to the Lab should now again redirect to the API login page

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Closes azavea/climate-change-api#732


fungjj92:
Custom CORS isn't working. I set my lab to another port (not 4200) and requests to /api-token still succeeded, sometimes.

screen shot 2017-11-08 at 1 31 03 pm
screen shot 2017-11-08 at 1 31 11 pm

PR #324, opened 2017-11-28
Fix tslint whitespace issue in Jenkins env

CloudNiner:
## Overview

tslint throws a "missing whitespace" error in the jsdoc comment of image-export.service.ts, but only in the jenkins environment. 

This issue is tracked here: https://github.com/palantir/tslint/issues/3251 and a scan through the comments indicates a general :shrug: that its caused by some magic combination of node + npm+  yarn + typescript + tslint + tsutils and that updating to yarn 1.2+ fixes it. I attempted this upgrade in place while SSHed into Jenkins but had no luck. I might have done something wrong getting node_modules cleaned out though.

So, for now I just removed the `whitespace.check-separator` in tslint as a workaround. I'm hopeful that a separate task to update yarn in the docker container on Jenkins would address the deeper issue. ~~Edit: first attempting a yarn version upgrade on fresh Jenkins build.~~

Created #325 to more thoroughly fix the issue.

## Testing Instructions

Jenkins should now build successfully.

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

### DIFF
--- a/src/app/services/image-export.service.ts
+++ b/src/app/services/image-export.service.ts
@@ -20,8 +20,8 @@ export class ImageExportService {
     /**
      * Converts chart SVG to PNG and downloads it.
      *
-     * @param indicatorName {string} Name of indicator; used for SVG selector
-     * @param fileName {string} File name for download; will be suffixed with extension
+     * @param indicatorName - Name of indicator, used for SVG selector
+     * @param fileName - File name for download, will be suffixed with extension
      */
     public downloadAsPNG(indicatorName: string, fileName: string): void {
         const filename: string = fileName + '.png';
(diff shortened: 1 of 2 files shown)


fungjj92:
Now that Jenkins is passing unexpectedly without this PR -- begs the question now, is this even necessary? Is Jenkins fooling us? Is it a different issue? 😥

CloudNiner:
Yuck I know. I don't believe its necessary, I'd prefer to not have to disable a fairly useful linter check.  But, I have no idea what changed that would have fixed the problem. If jenkins remains stable throughout the day I'll just close this with a relevant comment on #325.

CloudNiner:
Since Jenkins unfixed itself after a short period of time, I'm going to merge this so that builds return to a consistent state. We can address the root cause in the followup issue at a later time.

PR #334, opened 2019-02-13
Update most dependencies

### LINKED ISSUE #295: Update dependencies
Check for newer versions of package dependencies and upgrade if possible.

Things to check include:
  - npm packages for the front end
  - ansible role versions
  - installed system packages, such as node

ddohler:
## Overview

Updates to the latest minor version of most dependencies.

### Notes

- Angular, rxjs, and typescript could not be updated due to type conflicts with `climate-change-components`, and will need to be updated as part of, or after, updating `climate-change-components` to the latest version. 
- I wasn't able to run `yarn run lint` but it looks like that's because it's missing a configuration file that I'm not sure where to find (`The specified path does not exist: 'e2e/tsconfig.e2e.json'.`). I'm looking for help resolving that, and I'll update the README with whatever I do. --Edit-- This is also causing CI to fail. I'm not really sure about where to get started with this; it might be necessary to pair if there's not a quick and easy answer.
- Also includes an update to the README, because it looks like the app doesn't use an API key--it just redirects to the Climate Change API to set a cookie.

## Testing Instructions

 * `yarn install`
 * `yarn run X` everything in the `scripts` key of `package.json` and confirm there are no type errors (a few of them failed for reasons that looked unrelated to dependency upgrades to me).
 * Point your `src/app/constants.ts` to `http://localhost:8080` and bring up your Climate Change API VM. Click around in the lab and make sure there are no console errors and everything works. I don't know the Lab interface very well, so it is very possible I missed some functionality that could have been broken by one of these updates; thorough testing here will be appreciated.

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean? <-- I wasn't able to get this to run, see above.

Closes #295


package.json
```diff
@@ -32,45 +32,45 @@
...
-    "karma-coverage-istanbul-reporter": "^1.3.0",
-    "karma-jasmine": "~1.1.0",
+    "karma-coverage-istanbul-reporter": "^1.4.3",
+    "karma-jasmine": "~1.1.2",
     "karma-jasmine-html-reporter": "^0.2.2",
-    "protractor": "~5.1.2",
+    "protractor": "~5.4.2",
     "ts-node": "~3.3.0",
-    "tslint": "~5.7.0",
+    "tslint": "~5.12.1",
```
maurizi: By reverting the update to `tslint` I was able to get linting to pass again. The `e2e/` directory is used for end-to-end tests, which this project doesn't have. It looks like the support for `e2e` was somewhat removed, but not enough to make the latest version of `tslint` happy. I briefly tried removing further references to `e2e` but I wasn't able to get the new `tslint` to stop complaining about the missing file.

ddohler: :+1: Pushed up a commit to revert the tslint upgrade. Thanks for the help with this!

maurizi (approved):

PR #337, opened 2019-02-19
Update Angular to latest point release

ddohler:
## Overview

Updating Angular (and associated packages) wasn't possible as part of #334 because the `climate-change-components` dependency needed to be updated first. Now that #336 is merged, this updates Angular to the latest point release.

### Notes

Upgrading Typescript to 2.6.2 (from 2.6.1) still caused `ERROR in AppModule is not an NgModule`. The suggested resolutions I've found all revolve around installing particular Typescript versions, or particular Angular versions; I haven't been able to find a root cause for the issue, which seems to have cropped up across many combinations of Typescript and Angular versions. It didn't seem worth the effort to dig into what's actually going on, so I think we should let that one sit, until we either need one of the [bug fixes in Typescript 2.6.2](https://github.com/Microsoft/TypeScript/issues?q=is%3Aissue+milestone%3A%22TypeScript+2.6.2%22+label%3A%22fixed%22+), or until we embark on a major-version upgrade.

## Testing Instructions

 * `yarn install`
 * `yarn run serve`
 * Make sure the frontend still functions properly

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Finishes #295 (already closed)


ddohler:
The issue with CI appears to be that Chromium has recently been updated. On my development box, I had Chromium 71 installed inside the app container already, and `/scripts/test` worked, but after I cleared out my existing images and rebuilt, Chromium 72 was installed and at that point I was able to replicate the segfaults from CI.

I started playing with the base image to see if that would have any effect, and I was able to get the build to pass by switching to the `node:6-jessie` from `node:6-slim` (which uses Debian Stretch), but that's because Jessie installs Chromium 57, which is much older.

So I think this is probably some kind of regression in Chromium. I'm inclined to let it sit on Jessie for the time being, and we can make an issue to try switching back to Stretch later. Open to other suggestions though!

maurizi (approved):

PR #348, opened 2019-03-29
Use ExportServices from climate-change-components

ddohler:
## Overview

Brief description of what this PR does, and why it is needed.

### Demo

extreme_cold_events_Philadelphia_NEX-GDDP_RCP85

### Notes

- As part of trying to diagnose some difficulties with compiling using a local version of `climate-change-components`, I discovered that the `rxjs` import migration here was incomplete. Because `climate-change-components` imports `Observable` the wrong way, it allows things to work that won't work once `climate-change-components` imports `Observable` correctly. I didn't end up merging the changes to `climate-change-components` that triggered these failures (precisely because they caused unexpected problems in other places and I didn't want to risk breaking unrelated functionality in Temperate and the Lab), but the additional imports here are (apparently) harmless and they prevent us from having to redo the work whenever https://github.com/azavea/climate-change-components/issues/49 gets resolved.
- I noticed as part of this that PNG download is apparently broken on Staging -- I think there was a regression in the current version of the PNG-saving library; the latest version seems to fix the problem.
- This is WiP until https://github.com/azavea/climate-change-components/pull/51 is merged and a new version is released. After that happens, this should be updated to pull in the latest version in `package.json`, and remove the dependencies for SVG export and file saving.

## Testing Instructions

 * Check out the code from https://github.com/azavea/climate-change-components/pull/51 in your development environment for that project.
 * In the CC components repo, run `yarn run build:library`
 * From this repo, run `yarn install`
 * From this repo, run `rsync -avrc --delete ../climate-change-components/lib-dist/ node_modules/climate-change-components/lib-dist/` <-- Substitute the path to your `climate-change-components` repo if that isn't correct.
 * Fire up `yarn run serve` and confirm that you can still download PNGs of charts and that you get no compilation errors.
 * Also confirm that you can still download JSON exports.

## Checklist
- [x] `yarn run serve` clean?
- [x] `yarn run lint` clean?

Connects https://github.com/azavea/temperate/issues/1171


package.json
```diff
@@ -49,7 +49,7 @@
     "ngx-clipboard": "^8.1.4",
     "nouislider": "^10.1.0",
     "rxjs": "^5.5.11",
-    "save-svg-as-png": "^1.4.9",
+    "save-svg-as-png": "^1.4.12",
```
maurizi: Should this dependency continue to be here now that it is also in `climate-change-components`?

Similarly, can we remove the dependency on `file-saver` now?

ddohler: Yep, I'm planning to remove this before merging; it'll be superfluous once https://github.com/azavea/climate-change-components/pull/51 is merged, but it's potentially necessary for testing in the meantime. I wasn't sure whether `file-saver` might be used in other places, but if you don't think it is then I can remove that as well.

ddohler:
This is no longer WiP and is ready for review; cc-components 0.4.0 [has been released](https://www.npmjs.com/package/climate-change-components), and I've updated this to pull in the new dependency.

maurizi (approved):

